PR #45, opened 2020-02-05
feat+docs: add readSummarizedExperiment

cvanderaa:
Split readFeatures into 2 separate functions. readSummarizedExperiment is called within readFeatures. The documentation is adapted to include both functions in the same help page.


R/Features-constructors.R
```diff
@@ -3,8 +3,20 @@
 ##' @description
 ##'
 ##' Convert tabular data from a spreadsheet or a `data.frame` into a
-##' `Features` object.
-##'
+##' `Features` object (`readFeatures`) or a `SummarizedExperiment` 
+##' (`readSummarizedExperiment`).
+##'
+##' @usage
```
lgatto: Don't hard-code the usage section here - it will be generated automatically by roxygen in the final Rd file.

R/Features-constructors.R
```diff
@@ -25,33 +37,34 @@
...
 ##' @import SummarizedExperiment
 ##'
-##' @seealso The [Features] class for an example on how to manipulate use
+##' @seealso The [Features] class for an example on how to use
 ##'     `readFeatures` and how to further manipulate the resulting data.
 ##'
 ##' @importFrom methods new validObject
 ##'
 ##' @md
+##' @aliases readFeatures
```
lgatto: If the man page is about `readFeatures`, then you need to add `readSummarizedExperiment` as an alias. Otherwise it's ok.

R/Features-constructors.R
```diff
@@ -88,7 +101,11 @@ readFeatures <- function(table, ecol, fnames, ..., name = NULL)  {
     }
     se <- SummarizedExperiment(assay,
```
lgatto: No need to create `se` to then return it right away. Just use the call to the constructor as last expression.

lgatto (changes requested):

cvanderaa:
Thanks a lot for the review ! Here are some updates and additional unit tests !

tests/testthat/test_readFeatures.R
```diff
@@ -26,6 +26,28 @@ test_that("readFeatures", {
...
     expect_error(readFeatures(f, ecol = ecol, name = "psms"))
-    expect_error(readFeatures(x, ecol = 1:10, name = "psms", fname = "not_present"))
-    expect_error(readFeatures(f, ecol = 1:10, name = "psms", fname = "not_present"))
+    expect_error(readFeatures(x, ecol = 1:10, name = "psms", 
+                              fname = "not_present"))
+    expect_error(readFeatures(f, ecol = 1:10, name = "psms", 
+                              fname = "not_present"))
 })
+
+test_that("readFeatures", {
```
lgatto, line 35: This unit test should be called `readSummarizedExperiment`.

lgatto:
Could you also add yourself as a contributor (`role = "ctb"`) to the DESCRIPTION file.

cvanderaa:
Thanks for including me in the package development!

lgatto (approved):

PR #48, opened 2020-02-18
Extended VariableFilter (docs+feat)

cvanderaa:
* Added a `not` parameter to `VariableFilter` allowing to negate the
filter
* Added a `contains` internal function that implements the match
function when `condition == "contains"`
* Added examples and unit tests for the case when using the
`"contains"` condition and when using the `not` parameter is used
to negate the filter.
* Changed the R version requirement from `3.5.3` to `3.5.0`


R/Features-filter.R
```diff
@@ -81,6 +85,9 @@
 ##'
 ##' ## using the formula interface
 ##' filterFeatures(feat1, ~ pval <= 0.03)
+##' 
+##' ## you can also remove all p-values that are NA (if any)
+##' filterFeatures(feat1, ~ !pval %in% NA)
```
lgatto: Is it possible to use something like `!is.na(pval)`?

R/Features-filter.R
```diff
@@ -60,14 +60,18 @@
...
+##' filterFeatures(feat1, VariableFilter("location", "itochon", "contains"))
 ##'
 ##' ## ----------------------------------------------------------------
 ##' ## Filter all features that aren't marked as unknown (sub-cellular
 ##' ## location) in the feature variable
 ##' ## ----------------------------------------------------------------
 ##'
 ##' ## using a user-defined character filter
-##' filterFeatures(feat1, VariableFilter("location", "unknown", condition = "!="))
+##' filterFeatures(feat1, VariableFilter("location", "unknown", condition = "!="))##'
```
lgatto, line 74: Could you remove the `##'` at the end of the line.

R/Features-filter.R
```diff
@@ -94,6 +101,8 @@
 ##' filterFeatures(feat1, VariableFilter("foo", "bar"))
 ##'
 ##' filterFeatures(feat1, ~ foo == "bar")
+##'
```
lgatto, line 104: Remove the two empty `##'` lines

R/Features-filter.R
lgatto, line 90: Is it possible to use something like `!is.na(pval)`?

R/Features-filter.R
```diff
@@ -189,3 +207,13 @@ filterFeaturesWithFormula <- function(object, filter, ...) {
                   })
     object[sel, , ]
 }
+
+
+## Internal function called by `filterFeaturesWithAnnotationFilter` when 
+## `condition` is `"contains"`
+contains <- function(x, value){
```
lgatto, line 214: ~~If this is an internal function, could you rename if `.contains`.~~

lgatto (changes requested):

R/Features-filter.R
cvanderaa: If I'm not mistaken, the user then has to specify `condition == ".contains"` which is not accepted by `AnnotationFilter`. 
* Maybe I should export it with some documentation like in `?base::startsWith`. I just think it's weird that `Features` would export a function that has nothing to do with MS.
* Maybe I should create an issue in `AnnotationFilter` or ask @jorainer ?

lgatto: @jorainer - could you help us with the `contains` condition. How does the `contains` and `%like%` work, where are they implemented. Is it normal that we need to implement our own `contains` function?

jorainer: The `AnnotationFilter` package does not do any filtering itself, it just defines the filter classes and allowed comparisons (basically it defines the API for the filtering).

The package using the filter has to implement all the logic to perform the filtering. Example `ensembldb`: I'm supporting `AnnotationFilter`s in that package and you can either pass them to the `genes`, `transcripts` ... function with the parameter `filter`, or you can use the `filter` method on the database itself. Internally, I'm converting the filter into the SQL query to extract data from the database. I have a method `ensDbQuery` that does this conversion (https://github.com/jorainer/ensembldb/blob/master/R/Methods-Filter.R) and uses internal functions in https://github.com/jorainer/ensembldb/blob/master/R/functions-Filter.R to actually *translate* the filter expressions to SQL.

Summarizig, I guess you will have to implement your own filter functions, and along with that you have to implement the way how you will want to deal with `"contains"` - the `%like%` is actually an SQL command - you will most likely want to use a `grepl` or similar (if you're doing string comparisons in plain R).

lgatto (approved):

PR #61, opened 2020-04-08
Issue58: fix AssayLinks when parent assay are removed from the `Features` object

cvanderaa:
When a parent assay is removed from a `Features` object (e.g. after subsetting), the `AssayLink` of the child assay is replaced by an empty `AssayLink` object. 

Related unit tests are added.


cvanderaa:
Sorry wrong branch...

PR #62, opened 2020-04-08
Issue58: fix child AssayLinks when parent assay is removed

cvanderaa:
When a parent assay is removed from a Features object (e.g. after subsetting), the AssayLink of the child assay is replaced by an empty AssayLink object.

Related unit tests are added.

### DIFF
--- a/R/Features-class.R
+++ b/R/Features-class.R
@@ -214,8 +214,17 @@ setMethod("show", "Features",
 ##' @exportMethod [
 setMethod("[", c("Features", "ANY", "ANY", "ANY"),
           function(x, i, j, ..., drop = TRUE) {
+              ## Subset the assays
               ans <- callNextMethod(x, i, j, ..., drop)
+              ## Subset the AssayLinks
               ans@assayLinks <- ans@assayLinks[names(ans)]
+              ## Removed lost links
+              allist <- lapply(ans@assayLinks, function(al) {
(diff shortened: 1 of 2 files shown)


R/Features-class.R
```diff
@@ -214,8 +214,17 @@ setMethod("show", "Features",
 ##' @exportMethod [
 setMethod("[", c("Features", "ANY", "ANY", "ANY"),
           function(x, i, j, ..., drop = TRUE) {
+              ## Subset the assays
               ans <- callNextMethod(x, i, j, ..., drop)
+              ## Subset the AssayLinks
               ans@assayLinks <- ans@assayLinks[names(ans)]
+              ## Removed lost links
+              allist <- lapply(ans@assayLinks, function(al) {
+                  if(!al@from %in% names(ans)) al <- AssayLink(name = al@name)
```
lgatto, line 223: Space between `if (`! :-)

lgatto: Also, no need to assign `al` and then return is immediately. Clearer to return `AssayLink()` right away.

cvanderaa: Right ! Is it sufficient to put only the `if` statement within the function or should I also explicit the return value if the condition is not met, e.i.:
```r
if (!al@from %in% names(al)) {
    AssayLink(name = al@name)
} else {
    al
}
```

lgatto: Sorry, I misread the code. It is correct as it is.

PR #74, opened 2020-05-26
fix: allow subsetting of 'Features' objects containing 'SingleCellExperiment' assays

cvanderaa:
see [issue 273](https://github.com/waldronlab/MultiAssayExperiment/issues/276) in
'MultiAssayExperiment' for background

### DIFF
--- a/R/subsetBy-methods.R
+++ b/R/subsetBy-methods.R
@@ -110,8 +110,11 @@ setMethod("subsetByFeature", c("Features", "character"),
         i <- featurename_list[[k]]
     }
     
-    expts <- experiments(x[featurename_list, ])
-    alnks <- x@assayLinks[all_assays_names]
+    ord <- order(match(names(featurename_list), names(x)))
+    featurename_list <- featurename_list[ord]
+    expts <- subsetByAssay(x, names(featurename_list))
+    expts <- experiments(subsetByRow(expts, featurename_list))
+    alnks <- x@assayLinks[names(featurename_list)]
     alnks <- alnks[featurename_list]
 
     Features(experiments = expts,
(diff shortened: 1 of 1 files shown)


cvanderaa:
Wait before merging, it seems I inserted a bug...

cvanderaa:
In my opinion, the `subsetByFeature` has a strange behavior that causes the latest commit to break. See here (tested with the version of `Features` before ea8185d):
```r
data("feat1")
feat1 <- aggregateFeatures(feat1, 1, fcol = "Sequence", 
                           name = "peptides", fun = colMedians)
feat1 <- aggregateFeatures(feat1, 2, fcol = "Protein", 
                           name = "proteins", fun = colMedians)

names(feat1)
## [1] "psms"     "peptides" "proteins"
names(feat1["ProtA"])
## [1] "proteins" "peptides" "psms"    
```
@lgatto, don't you think subsetting features should preserve the assay ordering? 

**Explanation of the error:** The last commit replaces `expts <- experiments(x)[featurename_list]` with `expts <- experiments(x[featurename_list, ])`. The former subsets the experiments according to the ordering of the names in `featurename_list`, the latter subsets the experiments from the list but keeping the ordering of the assays in `x`. `AssayLinks` are subset according to the ordering in `featurename_list`. Hence in the last commit, the `AssayLinks` object is not ordered as the assays in `x`. This is detected as faulty by the `validObject`. So the solution is to either reorder the assays (lose the original assay ordering) or reorder the `AssayLinks` (preserve the original assay ordering)

lgatto:
@cvanderaa as you can see, that change lead to some unit tests failing. Could you look into this and identify why the tests now fail.

cvanderaa:
**Warning**: the unit tests will fail when PR #74 and PR #77 will be both merged into `master`. This is because the reordering discussed [above](https://github.com/rformassspectrometry/Features/pull/74#issuecomment-633926970) changes the expected output values in `test_subsetByFeatures.R`.

Issue #75, opened 2020-05-26
Count number of peptides in aggregateFeatures()


lgatto:
Following up from a discussion with @statOmics, the `msqrob` hurdle model requires peptides counts to assess whether a feature is differentially detected if it can't be tested for differential expression. 

It would be useful to all to calculate the number of peptides per protein per sample as part of the `aggregateFeatures()`. It remains to be seen whether this peptide count matrix is to be stored as an additional assay in the aggregated `SummarizedExperiment` or in the `rowData`.

lgatto:
In addition, irrespective where the counts matrix ends up, it will be useful to have an accessor - for example ~~`aggregationCounts()`~~ `aggcounts()`.

statOmics:
The accessor is a great idea. With an accessor it is not so important for our developments where the count matrix ends up...

lgatto:
@statOmics - see PR #76. Expect it to become available tomorrow, which requires `MsCoreUtils` 1.1.2 (all currently a [PR](https://github.com/rformassspectrometry/MsCoreUtils/pull/54)).  

The accessor is called `aggcounts()`, to be in line with `SingleCellExperiment::counts()`, `SingleCellExperiment::logcounts()`, ...

lgatto:
A quick vote here to name the aggregation count accessor. Currently, we have

1. **`aggcounts()`**: short, in line with other existing assay accessors such as `SingleCellExperiment::counts()`, `SingleCellExperiment::logcounts()`, ... but does sound like [*aggressive counts*](https://github.com/rformassspectrometry/Features/pull/76#discussion_r431610034) for previous discussion)).
2. **`aggregationCounts()`**: explicit, longer, no aggressive tone.

- To vote for 1: :+1: 
- To vote for 2: :-1: 
- Other suggestion: comment below and upvote that suggestion.

Ping @jorainer, @statOmics, @cvanderaa, @sgibb

PR #76, opened 2020-05-27
feat: new aggcounts with aggregation counts (see issue #75)

lgatto:


lgatto:
- @cvanderaa: could you have a look and check if everything looks good for you
- @jorainer: feel free to also have a look; this new feature related to feature aggregation might be of interest to you too.

NAMESPACE
```diff
@@ -24,6 +24,7 @@ exportClasses(Features)
 exportClasses(NumericVariableFilter)
 exportMethods("[")
 exportMethods("names<-")
+exportMethods(aggcounts)
```
jorainer, line 27: Suggestion - what about calling that function `featureCounts` instead? The first thing that came to my mind when reading `aggcounts` was *aggressive counts* ;)

cvanderaa:
It looks good to me. I really like the fact that `aggcounts` matrix is stored in the assay of the `SummarizedExperiment` object.

R/Features-aggregation.R
```diff
@@ -114,40 +119,56 @@
...
 ##' ## By default, missing values are propagated
-##' assay(aggregateFeatures(ft_na, 1, fcol = "X", fun = colSums), 2)
-##'
+##' ft2 <- aggregateFeatures(ft_na, 1, fcol = "X", fun = colSums)
+##' assay(ft2[[2]])
+##' aggcounts(ft2[[2]])
+##'
+##' ## The rowData .n variable tallies number of initial rows that
+##' ## were aggregated (irrespective of NAs) for all the samples. It
+##' #is always >= that the sample-level aggcounts.
```
cvanderaa: there is a missing `#<whitespace>`. Should the insertion of `.n` in the  `rowData` also be reported in the documentation, perhaps the `Details` section?

tests/testthat/test_aggregateFeatures.R
```diff
@@ -110,3 +110,20 @@ test_that("aggregateFeatures(fun = median)", {
...
+    ## assay
+    m <- matrix(c(NA, NA, NA, 14, 20, 22), nrow = 2)
+    rownames(m) <- 1:2
+    colnames(m) <- LETTERS[1:3]
+    expect_identical(assay(se), m)
+    ## aggcounts
+    m <- matrix(c(1, 1, 1, 2, 2, 2), nrow = 2)
+    rownames(m) <- 1:2
+    colnames(m) <- LETTERS[1:3]
+    expect_identical(aggcounts(se), m)
```
cvanderaa, line 128: We could also add a test on the `.n` variable stored in the rowdata. For example:
```r
## .n in rowData
expect_identical(rowData(se)$.n , c("1" = 2L, "2" = 2L))
```

NAMESPACE
lgatto: I initially planned for `aggregationCounts()` because it is called as part of the aggregation function. But given that there is already `SingleCellExperiment::counts()`, `SingleCellExperiment::logcounts()`, `SingleCellExperiment::tpm()`, ... I went for something short and all lower case.  

Open for discussion...

R/Features-aggregation.R
lgatto: Good points. Will be added.

tests/testthat/test_aggregateFeatures.R
lgatto: Added, thanks.

lgatto:
This PR addresses issue #79. It still misses unit test though.

PR #77, opened 2020-06-04
Issue52: adding AssayLinks with multiple parents

cvanderaa:
Implemented new functionality to add `AssayLinks` linking a child assay to multiple parents. 

This required the modification of the `AssayLink` signature: the `hits` slot can now contain `Hits` or `List` objects, and the `from` slot can be a vector. The `AssayLinks` validity checks and the `show` method have been updated to account for the new signature.

The `joinAssays` was adapted so that it automatically creates a multi-parent assay links between the assays to join and the joined assay. `subsetByFeatures` method and the `"["` method for `Features` have also been adapted to deal with multiple parent assays.

Documentation and unit tests were also added.


R/AssayLinks.R
```diff
@@ -257,6 +272,52 @@ setMethod("[", c("AssayLinks", "list"),
...
+        rdTo <- cbind(rowData(object[[to]]), RowNames = rownames(object[[to]]))
+        hitsList <- lapply(seq_along(from), function(ii) {
+            rdFrom <- cbind(rowData(object[[from[ii]]]), 
+                            RowNames = rownames(object[[from[[ii]]]]))
+            .create_assay_link(rdFrom = rdFrom,
+                               rdTo = rdTo,
+                               from[ii], to,
+                               "RowNames", "RowNames")@hits
+        })
+        varsFrom <- "RowNames"
```
cvanderaa: This conditional code chunk might seem dirty but is necessary for matching `rownames` to parent assays (only possible when called within `joinAssays`). Maybe you can think about something cleverer?

cvanderaa:
The check gives the following error (converted from warning).
```
* checking whether package ‘Features’ can be installed ... WARNING
Found the following significant warnings:
  Warning: multiple methods tables found for ‘which’
See ‘/home/travis/build/rformassspectrometry/Features/Features.Rcheck/00install.out’ for details.
```
I don't get this locally, any idea what this means?

cvanderaa:
I've updated the PR to include the comments of our last meeting. I'm done unless you think there are still changes required.

cvanderaa:
This is ready to be merged on my side.

Issue #105, opened 2020-09-04
feat: add `infIsNA`?


cvanderaa:
I have implemented an [`infIsNA`](https://github.com/UCLouvain-CBIO/scp/blob/92f78503ffc604ab36f637c3e38b43e52dfaec0c/R/transform.R#L96-L120) function in the `scp` package. It replaces any infinite value by `NA`. This occurs when dividing data by a reference sample that contains 0. 

I am wondering whether this function should be provided by `QFeatures` rather than `scp`. What do you think? 

I could convert the function to a method so that it behaves like `zeroIsNA`.

lgatto:
Yes, I think it would be a better fit for `QFeatures`, as it is very general.

PR #106, opened 2020-09-04
Issue105: add `infIsNA` 

cvanderaa:
Added a new feature, `infIsNA`, that replaces infinite values in one or more assays with `NA`. This was suggested in issue #105.


lgatto:
I am wondering if we wouldn't need a `infIsNA,SummarizedExperiment,integer` method for when a `SummarizedExperiment` object has multiple assays. I know this should never be the case for `QFeatures` object, and this that the integer would by default always be 1, but it would make sense to be able to use the method more widely with `SummarizedExperiment` objects.

This actually probably also applies to many other methods: `zeroIsNA`, ...

cvanderaa:
This could indeed be added but then I would also add the signatures `infIsNA,SummarizedExperiment,character`?

lgatto:
Do assays of a `SummarizedExperiment` have to be named? If not, then I would prioritise the integer for all the functions we have, and only add the character methods either if we identify a need or if it matches best practice.

cvanderaa:
They don't have to be named.

cvanderaa:
I realize there is an error inn the CI check: `dist: trusty is no longer supported for language: r`. Should we upgrade the `dist` to `xenial`?

lgatto:
I would suggest to move to github actions: see the [`scp/.github`](https://github.com/UCLouvain-CBIO/scp/tree/master/.github) directory for details.

cvanderaa:
I replaced the `yml` files to enable GitHub actions, but since nothing is happening I guess we have to add `actions-user` as a contributor. Or am I missing something else?

lgatto:
May be it needs to get merged to master for the actions to kick in?

cvanderaa:
Indeed thanks!

Issue #108, opened 2020-09-09
'colnames' are lost when reading in column with 'readSummarizedExperiment'


cvanderaa:
See example: 
```r
library(QFeatures)
data("hlpsms")
```
Normal behavior for reading in multiple quantification 
```r
test1 <- readSummarizedExperiment(hlpsms, ecol = 1:10)
colnames(test1)
```
```
[1] "X126"  "X127C" "X127N" "X128C" "X128N" "X129C" "X129N" "X130C" "X130N" "X131" 
```
`colnames` are lost when reading in a single column
```r
test2 <- readSummarizedExperiment(hlpsms, ecol = 1)
colnames(test2)
```
```
NULL
```

cvanderaa:
Haha nice, I see you went faster than me! You can ignore PR #109.

PR #109, opened 2020-09-09
Fix Issue #108: preserve column names when reading 1 column 'SummarizdExperiment'

cvanderaa:
Also included unit tests.


lgatto:
Re-opening because you added a unit test - I owe you a drink. Will merge and fix conflicts.

Issue #110, opened 2020-09-10
'QFeatures' subsetting leads to wrong 'AssayLinks' when multiple parents


cvanderaa:
See the example taken from `?addAssayLink` with multiple parents: 
```r
se <- SummarizedExperiment(matrix(runif(20), ncol = 2,
                                  dimnames = list(LETTERS[1:10],
                                                  letters[1:2])),
                           rowData = DataFrame(ID = 1:10))
ft <- QFeatures(list(parent1 = se[1:6, ], parent2 = se[4:10, ], child = se))
ft <- addAssayLink(ft, from = c("parent1", "parent2"), to = "child",
                   varFrom = c("ID", "ID"), varTo = "ID")
```
The issue arises after subsetting features in a `QFeatures` object for the assay that is linked to multiple parents, `child` in this case. While no errors are thrown, see here the bug:
```r
ftsub <- subsetByFeature(ft, "A")
assayLink(ftsub, "child")
```
```
AssayLink for assay <child>
[from:parent1,parent2|fcol:ID,ID|hits:]
```
`hits` is empty!

This can be solved by adapting `"[", c("AssayLink", "character")` method. I'll make a PR soon, but suggest to add an additional validity check.

PR #111, opened 2020-09-10
Issue #110: fix `AssayLink` subsetting for multiple parents

cvanderaa:
See issue #110 for background information. 

Note: I added a validity check on `QFeatures` objects. When implementing this check without solving the bug makes the `test_subsetByFeatures.R` fail indicating that the issue is now automatically identified by the unit tests.


cvanderaa:
Wait before merging...

cvanderaa:
Ok I'm good on the implementation part, but some checks fail on Windows and macOS because the `biocversion` is not correctly supplied. Any idea how to avoid this?

Issue #86, opened 2020-07-07
Function to extract all row data


lgatto:
See `scp::rowDataToDF()`.

cvanderaa:
The current `scp::rowDataToDF` function signature is:
- `object` A `QFeatures` object
- `i` A numeric() or character() vector indicating from which assays the `rowData` should be taken.
- `vars` A character() vector indicating which variables from the `rowData` should be extracted.

**To discuss**

1. I don't like the naming, since `rowData` should always return a `DataFrame`. What do you think about calling the function `collectRowData`, as it boils down to *picking* the `rowData` tables from different assays and put them in the same table (or *basket* if we stick to the gardening analogy :laughing:). An alternative would be `extractRowData`. 
2. **(edited)**: I'm wondering whether the `vars` is really needed, or at least we could implement that when missing, the function keeps all `rowData` variables *that are common to all selected assay*. 
3. Similarly, we could implement that the `rowData` of all assays are collected when `i` is missing.
4. Rename `vars` to: (1) `rowDataCols` for consistency with, *e.g.*, `MultiAssayExperiment::longFormat` or (2) `rowvars` for consistency with `selectRowData`.

@lgatto what do you think?

lgatto:
I have seen your PR but haven't looked at it. 

My preference would be to use `rowData`, which is already known, and return a list of row data. However, the `...` in the signature is ignores, which means that the parameters you suggest can't be used. I don't think this is a problem because:
- The `i` you suggest isn't necessary, as it is possible to filter the `QFeatures` object to keep the assays of interest before extracting to row data list.
- The `vars` could be useful when you only want the same variables acorss all assays, but I don't think it is essential, as you mention.

If we have the same variable across all assays, then that list could be converted to a DataFrame with `do.call(rbind, x)`. If the variables are different, then `MsCoreUtils::rbindFill` should do the trick (not tested).

cvanderaa:
This makes sense indeed. I'll adapt my PR.

cvanderaa:
I think this issue can be closed unless you have additional suggestions?

PR #115, opened 2020-09-17
Issue86: implement a function to collect rowData from a `QFeatures` object

cvanderaa:
See issue #86 for background. I took the following decision regarding the points that were addressed in the issue:

- I renamed the function from `scp::rowDataToDF` to `collectRowData`
- I named the argument `rowvars` that allows the user to select the columns in the `rowData`.
- If `rowvars` is missing, the shared `rowData` variables between the selected assay will be taken. If none exist, an error is thrown
- If `i` is missing, the `rowData` is extracted from all assays


cvanderaa:
The `collectRowData` function is now replaced by a `rowData` method on `QFeatures` objects.

PR #119, opened 2020-10-23
Issue118: fix nNA

cvanderaa:
Fixed the nNA method and added a fow more unit tests.


lgatto:
@cvanderaa - could you bump de version to 0.99.4 and update the news file accordingly, please.

lgatto:
I'll do it.

Issue #125, opened 2021-02-08
Check fails on Bioconductor servers


lgatto:
```
* checking tests ...
  Running ‘testthat.R’
 ERROR
Running the tests in ‘tests/testthat.R’ failed.
Last 13 lines of output:
  
  `actual`:   FALSE
  `expected`: TRUE 
  ── Failure (test_subsetByFeatures.R:41:5): subsetByFeatures ────────────────────
  `res1` not equal to `res2`.
  Attributes: < Component "drops": names for current but not for target >
  Attributes: < Component "drops": Length mismatch: comparison on first 0 components >
  ── Failure (test_subsetByFeatures.R:56:5): subsetByFeatures: full pipeline ─────
  `ftsub` not identical to fts["ProtA", ].
  Attributes: < Component "drops": names for current but not for target >
  Attributes: < Component "drops": Length mismatch: comparison on first 0 components >
  
  [ FAIL 8 | WARN 37 | SKIP 0 | PASS 351 ]
  Error: Test failures
  Execution halted
```
All details here:  http://bioconductor.org/checkResults/devel/bioc-LATEST/QFeatures/

lgatto:
Ping @cvanderaa - could you have a look

cvanderaa:
I'll have a look. I suspect this has to do with the recent changes in `MultiAssayExperiment`.

cvanderaa:
So the issue is linked to updates in MAE. This means that as soon as we perform subsetting on the experiments, a warning is thrown: `'experiments' dropped; see 'metadata'` and the dropped experiments are listed in the metadata slot. I think this can lead to quite some confusion in our package to users as this could make them believe they are doing something wrong. Maybe we should catch the warning and hide it from the user where this warning is expected. 

There are quite some warnings and errors that come out because of that, so maybe we should think about this before adapting the tests?

These changes are probably an answer to those issues: https://github.com/waldronlab/MultiAssayExperiment/issues/290 and https://github.com/waldronlab/MultiAssayExperiment/issues/289. Maybe we should contribute to the discussion?

lgatto:
I saw that Marcel committed a fix to this. `scp` should now depend on that specific version of `MultiAssayExperiment` by specifying the version in the Depends field.

cvanderaa:
[...]

PR #126, opened 2021-02-22
refactor: optimized 'infIsNA' and 'zeroIsNA'

cvanderaa:
I noticed that `infIsNA` and `zeroIsNA` replace assays in a loop what leads to calling `MultiAssayExperiment:::.harmonize` at each iteration, which is very slow (see also https://github.com/waldronlab/MultiAssayExperiment/pull/297). 

I now extract the `ExperimentList` and modify it in the loop (very fast) and then replace that slot of the `QFeatures` object with the modified list. `BiocGenerics:::replaceSlots` checks the validity of the object. 

Note: the speed could further be increased by setting `check = FALSE` in the `BiocGenerics:::replaceSlots`. Since we know by the design of `zeroIsNA, SummarizedExperiments` that nothing changes except the quantification tables, we know the returned object is valid.

### DIFF
--- a/R/QFeatures-missing-data.R
+++ b/R/QFeatures-missing-data.R
@@ -171,9 +171,12 @@ setMethod("zeroIsNA", c("SummarizedExperiment", "missing"),
 ##' @rdname QFeatures-missing-data
 setMethod("zeroIsNA", c("QFeatures", "integer"),
           function(object, i) {
+              el <- experiments(object)
               for (ii in i)
-                  object[[ii]] <- zeroIsNA(object[[ii]])
-              object
+                  el[[ii]] <- zeroIsNA(el[[ii]])
+              BiocGenerics:::replaceSlots(object,
+                                          ExperimentList = el,
+                                          check = TRUE)
           })
 
 ##' @rdname QFeatures-missing-data
@@ -183,9 +186,9 @@ setMethod("zeroIsNA", c("QFeatures", "numeric"),
 ##' @rdname QFeatures-missing-data
 setMethod("zeroIsNA", c("QFeatures", "character"),
           function(object, i) {
-              for (ii in i)
-                  object[[ii]] <- zeroIsNA(object[[ii]])
-              object
+              if (any(! i %in% names(object)))
+                  stop("subscript contains invalid names")
+              zeroIsNA(object, which(names(object) %in% i))
           })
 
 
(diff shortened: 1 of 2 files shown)


lgatto:
1. Have you made sure that these new implementations return the exact same results as the previous ones? 

2. I don't understand why `MultiAssayExperiment` is supposedly not available and the check fails; do you have an idea?

R/QFeatures-missing-data.R
```diff
@@ -183,9 +186,9 @@ setMethod("zeroIsNA", c("QFeatures", "numeric"),
 ##' @rdname QFeatures-missing-data
 setMethod("zeroIsNA", c("QFeatures", "character"),
           function(object, i) {
-              for (ii in i)
-                  object[[ii]] <- zeroIsNA(object[[ii]])
-              object
+              if (any(! i %in% names(object)))
+                  stop("subscript contains invalid names")
```
lgatto, line 190: My latest compilation of the `SCP.replication` vignette failed with that message. Did you check `SCP.replication` with these changes?

cvanderaa:
1. Yes all checks related to `infIsNA` or `zeroIsNA` still pass. The issues that remain are linked to #125 
2. I noticed `MultiAssayExperiment` has built issue on GitHub, but don't know if that was pushed to the Bioconductor server...

lgatto:
> Yes all checks related to `infIsNA` or `zeroIsNA` still pass. 

Yes, they do, but that doesn't mean the outputs are identical. It would be interesting to check that they remain `identical()` - this wouldn't need to be a unit test that is run every time Sometimes, such verifications are helpful as they help detect some small hidden (and arguably insignificant) differences, but useful to understand some implementation details of the objects.

lgatto:
> I noticed MultiAssayExperiment has built issue on GitHub, but don't know if that was pushed to the Bioconductor server.

I don't know and don't understand. The latest version available should install from Bioconductor. Weird.

cvanderaa:
> It would be interesting to check that they remain `identical()`

True! I will double-check this. I also now realize that the unit tests could be improved!

cvanderaa:
I confirm the two implementations return identical object. I also improved the unit test.

lgatto:
Looks like this can also be merged - could you confirm.

cvanderaa:
I do confirm!

PR #127, opened 2021-02-22
fix: avoided warnings thrown by 'MultiAssayExperiment'

cvanderaa:
Starting to solve the identified issues as listed in https://github.com/rformassspectrometry/QFeatures/issues/125:

- [x] `test_Features.R`: we are checking the QFeatures subsetting method. MAE throws an error as soon as an assay is removed. I think we could simply catch the warning with no further change, what do you think?
- [ ] `test_filerFeatures.R` and `test_subsetByFeatures`: assays that contain 0 features or 0 columns are no longer removed from the object when performing subsetting with [. Some tests in test_filerFeatures.R are expecting empty objects but this does no longer happen. BUT subsetByFeatures does remove empty assays. There is a diverging behavior between [ and subsetByFeatures. What behavior should we keep for QFeatures? I would prefer the [ behavior (with a drop argument) since this is in line with MAE.
- [x] `test_join.R`: a warning is thrown because joinAssays performs an internal subsetting. This can be easily avoided with a little refactoring.
- [x] `test_missing-data.R`: can be easily avoided with a little refactoring in the tests.

### DIFF
--- a/R/QFeatures-join.R
+++ b/R/QFeatures-join.R
@@ -152,7 +152,7 @@ joinAssays <- function(x,
               "Need at least 2 assays to join" = length(i) >= 2)
     if (name %in% names(x))
         stop("Assay with name '", name, "' already exists.")
-    joined_se <- mergeSElist(as.list(experiments(x[, , i])))
+    joined_se <- mergeSElist(as.list(experiments(x)[i]))
     x <- addAssay(x, joined_se, name = name)
     ## Add the multi-parent AssayLinks
     if (is.numeric(i)) i <- names(x)[i]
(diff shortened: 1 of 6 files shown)


lgatto:
1. `test_Features.R`: we are checking the QFeatures subsetting method. MAE throws an error as soon as an assay is removed. I think we could simply catch the warning with no further change, what do you think?

Error or warning? I assume it's the latter that triggers an error in the unit test.

lgatto:
2. Yes, I think we should remain consistent with MAE.

cvanderaa:
> Error or warning? I assume it's the latter that triggers an error in the unit test.

Oops, you're right. I meant warning.

cvanderaa:
The tests should be fixed by now, but I still need to work on point 2. If you are ok with it, you can merge the PR and I'm opening a new issue.

PR #132, opened 2021-04-21
fix: fixed unexpected behavior when ecol is logical

cvanderaa:
I adapted `readSummarizedExperiment` and created a related unit test.


cvanderaa:
Build seems to fail, the build error is:
```
Error in gsub("\r\n", "\n", str, fixed = TRUE) : 
  input string 1 is invalid in this locale
Calls: <Anonymous> -> new_rcmdcheck -> win2unix -> gsub
Execution halted
```
I have no clue what this means...

PR #133, opened 2021-04-22
Add replaceRowDataCols and removeRowDataCols

cvanderaa:
Migration of the `replaceRowDataCols` from `SCP.replication` to `QFeatures`. I also added a `removeRowDataCols` function. 

Those functions are intended to facilitate manipulation of the `rowData` of assays in `QFeatures` object. I adapted the `QFeatures-class` documentation and added unit tests.

Before merging I would like to discuss two things:

- I should add an example on how to add, replace and remove rowData columns. Do you prefer to have this in the documentation (downside: the example section become crowded) or in the vignette (downside: not directly accessible by user).
- I'm not convinced by the naming `replaceRowDataCols` since the function is able to replace **and** add columns.


cvanderaa:
Note that this PR will be improved by PR #134

PR #139, opened 2021-06-30
Improved validity checks and solve subsetting issue

cvanderaa:
In an attempt to solve https://github.com/UCLouvain-CBIO/SCP.replication/issues/8, I realized that the validity checks on `AssayLinks` could be improved. I refactor the checks, put them in a separate file, and added unit tests. 

The unit tests for `filterFeatures` currently fails because it now detects an overlooked bug during subsetting. I'm working on it so **don't merge yet**. I still need to: 
- [x] Improve validity checks
- [x] Debug the unit tests for `filterFeatures`
- [x] Make sure that this solves https://github.com/UCLouvain-CBIO/SCP.replication/issues/8
- [x] Update `NEWS.md` and version bump


cvanderaa:
I finished the PR, but there seems to be an issue that I cannot reproduce locally, even with the latest packages...

lgatto:
It looks like this is the same error that we currently have in `MSnbase` (see [here](https://github.com/lgatto/MSnbase/runs/2969471507?check_suite_focus=true#step:12:32)):
```
attempt to set 'rownames' on an object with no dimensions
```

Ping @jorainer @sgibb in case you have an idea.

DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: QFeatures
 Title: Quantitative features for mass spectrometry data
-Version: 1.3.3
+Version: 1.3.4
```
lgatto: The version should stay 1.3.3, which corresponds to updates specific to github development, that will be pushed to Bioc devel (currently at 1.3.2).

NEWS.md
```diff
@@ -1,8 +1,13 @@
 # QFeatures 1.3.0
 
+## QFeatures 1.3.4
```
lgatto: See above.

lgatto (changes requested):

cvanderaa:
> It looks like this is the same error that we currently have in `MSnbase` (see [here](https://github.com/lgatto/MSnbase/runs/2969471507?check_suite_focus=true#step:12:32)):
> 
> ```
> attempt to set 'rownames' on an object with no dimensions
> ```
> 
> Ping @jorainer @sgibb in case you have an idea.

In both cases (`MSnbase` and `QFeatures`), the error appears during normalization.

cvanderaa:
Ah wait, I am able to reproduce the error locally using the GitHub version of `MsCoreUtils`!

cvanderaa:
The issue lies in `preprocessCore`, the `normalize.quantiles.robust` function is not behaving as expected. See this example:
```r
library(QFeatures)
library(preprocessCore)
data(hlpsms)
fts <- readQFeatures(hlpsms, ecol = 1:10, name = "psms")
se <- fts[[1]]
normalize.quantiles(assay(se))
## returns a 3010 by 10 matrix (as expected)
normalize.quantiles.robust(assay(se))
## returns NULL
```

lgatto:
could you post on issue on the MsCoreUtils repo, please.

lgatto:
Note that [MSnbase](http://bioconductor.org/checkResults/devel/bioc-LATEST/MSnbase/nebbiolo2-buildsrc.html) also fails on Bioconductor because of this.

lgatto:
See [here](https://github.com/bmbolstad/preprocessCore/pull/15) for the source of the error and a fix.

cvanderaa:
The PR now passes the checks, in my opinion it is ready for merge. When merging, would it be possible to push to Bioconductor? That would allow me to use the recent changes to solve the issues in `SCP.replication`.

lgatto:
Pushed to Bioc.

Issue #141, opened 2021-07-05
@hits contains links that point to missing features


lgatto:
From @statomics 
```r
pe <- normalize(pe, 
                i = "peptideLog", 
                name = "peptideNorm", 
                method = "center.median”)
```
produces
```
Error in .checkLinksInHits(al@hits, object[[al@name]], direction = "to") : @hits contains links that point to missing features
15. stop("@hits contains links that point ", direction, " missing features")
14. .checkLinksInHits(al@hits, object[[al@name]], direction = "to")
13. .validAssayLink(object, i)
12. .validAssayLinks(object)
11. validityMethod(object)
10. isTRUE(x)
9. anyStrings(validityMethod(object))
8. validObject(.Object)
7. initialize(value, ...)
6. initialize(value, ...)
5. new("QFeatures", ExperimentList = el, colData = colData(x), sampleMap = smap, metadata = metadata(x), assayLinks = append(x@assayLinks, assayLinks))
4. addAssay(object, normalize(object[[i]], method, ...), name)
3. .local(object, ...)
2. normalize(pe, i = "peptideLog", name = "peptideNorm", method = "center.median")
1. normalize(pe, i = "peptideLog", name = "peptideNorm", method = "center.median")
```

This happens with the latest github version, not with the one from Bioconductor. 

```SESSION INFO

R version 4.1.0 Patched (2021-05-23 r80364)
Platform: aarch64-apple-darwin20 (64-bit)
Running under: macOS Big Sur 11.2

Matrix products: default
LAPACK: /Library/Frameworks/R.framework/Versions/4.1-arm64/Resources/lib/libRlapack.dylib

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:
[1] parallel  stats4    stats     graphics 
[5] grDevices utils     datasets  methods  
[9] base     

other attached packages:
 [1] gridExtra_2.3              
 [2] plotly_4.9.4.1             
 [3] forcats_0.5.1              
 [4] stringr_1.4.0              
 [5] dplyr_1.0.7                
 [6] purrr_0.3.4                
 [7] readr_1.4.0                
 [8] tidyr_1.1.3                
 [9] tibble_3.1.2               
[10] tidyverse_1.3.1            
[11] msqrob2gui_0.0.1           
[12] openxlsx_4.2.4             
[13] ggplot2_3.3.5              
[...]

PR #142, opened 2021-07-06
Fixed bug that produced invalid AssayLinks when using filterNA (cf #141)

cvanderaa:
Fixed small bug in code so that `filterNA` now correctly subsets the `AssayLinks`.


lgatto:
Also pushed on Bioc.

Issue #152, opened 2021-11-18
feat: QFeatures needs to implement an `updateObject()` method


cvanderaa:
From a conversation with Hervé Pagès on Slack about whether to implement an `updateObject()` or not:

>No simple answer to that. It depends.

>1. First of all if the new class is for objects that are not meant to be serialized then updateObject() is not needed. For example nobody should ever serialize a TxDb object.
>2. Another situation where maybe there's no need to define a specific updateObject() method for a new class B is if the new class extends a class A that already has an updateObject() method, and adds slots that are meant to contain only ordinary vectors or other base R objects. Then calling updateObject() on objects of class B will do the right thing i.e. it will update the parts of the object that are under the control of A and will ignore the new slots. All this is fine until the author of class B decides to make changes to the class definition e.g. to add more slots. When this happens, they'll need to implement an updateObject() method for their objects (this method will need to call callNextMethod() and that call should happen at the very end of the method body).
>3. If, OTOH, B extends A by adding slots that are meant to contain other S4 objects, then it's recommended to define an updateObject() method for B. This method would simply update the new slots containing S4 objects by calling updateObject() on each of them, and call callNextMethod() at the end.
>4. If the new class does not extend anything then calling updateObject() on these objects will call the updateObject,ANY method defined in BiocGenerics. This default updateObject() method will probably do the right thing, until the author of the new class makes a change to the internal of their class. When this happens, they'll need to implement a specific updateObject() method for their class of course.

>In any case calling updateObject() on an S4 object is expected to work, even if it doesn't do anything (no-op). So I'd say that authors of a new class should make sure that updateObject() doesn't return an error on their objects.

I would say `QFeatures` belongs to 3. since its need to update the `@assayLinks` slot that expects an S4 object of class `AssayLinks`. Therefore we need to implement an `updateObject()` method to update the `@assayLinks`.

lgatto:
Indeed. Thank you for all this information!

cvanderaa:
Shall I work on it?

lgatto:
Yes, please. Thanks

cvanderaa:
Solved by #153

PR #153, opened 2021-11-19
Implmented an updateObject() method for QFeatures objects (issue #152)

cvanderaa:
This is my attempt to solve the updateObject() issue #152.

Although everything seems to work, I'm not 100% convinced of the implementation: https://github.com/cvanderaa/QFeatures/blob/a3550c0c0251766d1bffcb564eaadb0b95cd4151/R/QFeatures-class.R#L707-L722

Converting to and from an `MAE` is not elegant to me. An alternative I see is to pass the assayLinks as a `...` in `MultiAssayExperiment::updateObject()`, but that is not currently available on their side. I could ask Marcel Ramos for advice. What do you think?

PS: I realized I forgot to push a small change in the plot parameters. I believe this minor tweak does not deserve attention.


lgatto:
Does `callNextMethod()` trigger a call to `validObject()`? Otherwise, you could do 

```r
al <- object@assayLinks
object <- callNextMethod()
object@assayLinks <- al
```
because I agree that converting back and forward isn't great. 

I would ask on slack, in the same thread.

cvanderaa:
>Does callNextMethod() trigger a call to validObject()? 

Yes it looks like so... Could it be it is automatically triggered upon creation of the object with `new` ? https://github.com/waldronlab/MultiAssayExperiment/blob/4db4d7df99a4bc1cc209e9e7f36180235f1572d5/R/MultiAssayExperiment-class.R#L792-L805  

I tried the code you suggest, but it leads to the same initial error about wrong AssayLinks. That's why I would like to pass `al` as an argument to 
`object <- callNextMethod()`. My guess would be to add `...` at this line so that we can pass our desired slots:  https://github.com/waldronlab/MultiAssayExperiment/blob/4db4d7df99a4bc1cc209e9e7f36180235f1572d5/R/MultiAssayExperiment-class.R#L806
This would allow to call: 
```r
object <- callNextMethod(object, assayLinks = al, ..., verbose = verbose)
```
and avoid converting `object`. But this would require a change in `MAE`. I will ask on slack.

lgatto:
We don't want any change in `MAE`. 

Another option could be to call `updateObject(object, objclass = "MultiAssayExperiment")` to update the `MAE` part of our object, and then construct a new `QFeatures` and add the missing assay link.

cvanderaa:
I agree with you! I have modified the implementation using `objclass = "MultiAssayExperiment"`, it looks cleaner to me now! I'm good for the merge, unless you have additional suggestions.

lgatto:
Happy to hear this works. You might still check with the experts on Slack, to see what they suggest.

cvanderaa:
I refactored the method based on a discussion on the Bioc Slack. I also added `updateObject()` methods for `AssayLinks` and `AssaLink` objects. The 2 methods are currently very simple and serve rather as backbone for later class changes that may occur.

Issue #154, opened 2021-12-09
Add an adjacencyMatrix setter


lgatto:
- This  could also insure that we have a properly sparse matrix.
- We should also check that the adjacency matrix has row/col names

lgatto:
Done in latest PR.

PR #155, opened 2021-12-09
This PR enables aggregation by matrix on SEs and QFeatures

lgatto:
- `aggregateFeatures()` for SEs
- `aggregateFeatures()` for QFeatures
- AssayLink are created from an adjacency matrix
- A new `adjacencyMatrix()` setter (see #154) that converts the matrix to a sparse matrix and checks col/row names
- The  `adjacencyMatrix()` getter now also returns warning of the matrix isn't sparse and missing col/row names

TODO: 
- unit tests for the latest functionality, in particular `adjacencyMatrix()`


DESCRIPTION
```diff
@@ -33,6 +33,7 @@ Imports:
    plotly
```
cvanderaa, line 33: `MsCoreUtils (>= 1.1.2)` should become `MsCoreUtils (>= 1.7.2)`, otherwise `QFeatures` installation fails with `Error: object ‘aggregate_by_matrix’ is not exported by 'namespace:MsCoreUtils'`

lgatto: Indeed, thank you - addressed now.

Issue #104, opened 2020-08-16
addAssay: wrong colData when adding assay with new sample IDs


cvanderaa:
Suppose you have a `QFeatures` object and want to add an assay with new samples, using `addAssay` throws a warning. See for example:
```r
data("feat1")
se <- feat1[[1]]
## Change column names = new sample IDs
colnames(se) <- paste0("foo", seq_len(ncol(se)))
addAssay(feat1, y = se)
```
The last command throws a warning:
```
Warning message:
Data dropped from ExperimentList (element - column):
 newAssay - foo1
 newAssay - foo2
Unable to map to rows of colData 
```
The updated `colData` and `sampleMap` are wrong. For example, the `colData` is
```
DataFrame with 2 rows and 1 column
       Group
   <integer>
S1         1
S2         2
```
whereas we would expect
``` 
DataFrame with 4 rows and 1 column
       Group
   <integer>
S1         1
S2         2
foo1       1
foo2       2
```

cvanderaa:
An additional item is to remove redundant `colData` variables before adding the assay to the `QFeatures`. Whether the variables are removed or not could be controlled by an argument. It's the opposite of what is already performed in `getWithColData` (see [here](https://github.com/waldronlab/MultiAssayExperiment/blob/a26c8ff25bfe1cc9a2fdd578d502dbf2f81a9e46/R/MultiAssayExperiment-helpers.R#L631-L648))

cvanderaa:
closed by https://github.com/rformassspectrometry/QFeatures/commit/8563bdee2c107913dd9e1257556f081c34fc11ba

Issue #57, opened 2020-03-31
[[<-,Features


lgatto:
We will need to implement `[[<-,Features` that will
- check that the new assay has unique row names
- that the new row names match (possibly in a different order) to old ones, to still support `subsetByFeatures`

PR #165, opened 2022-07-20
Improve 'addAssay()' (#104)

cvanderaa:
I improved `addAssay()`. The improvement solves issue #104, but also dramatically improve computational speed.  

I also extended the unit tests to catch the error (and others) that is mentioned in issue #104.

Please, **do not merge**. There are important aspects I think need to be discussed:

- [x] Currently, the colData is automatically transferred from the assay to the QFeatures object. Do you like this behavior?
- [x]  An additional argument, `dropColData` controls whether to keep the colData in the assay that was added. It defaults to removing the colData for `addAssay()` to avoid redundant information and it defaults to keeping the colData for `replaceAssay()` otherwise `colData(x[[1]]) <- DataFrame(...)` does not work and would not allow users to play with the assay's colData. Is this what we want? 
- [x] When we add an assay that contains samples already in the QFeatures object and that has colData variables in common with the colData of the QFeatures, the data is overwritten. Again, is this something we want? (see [here](https://github.com/cvanderaa/QFeatures/blob/f5b5da58ba3723bac0190226f5e34d992461610a/tests/testthat/test_Features.R#L196-L204) for an example)


cvanderaa:
In parallel, I'll be working on an `replaceAssay()` and a `removeAssay()`. Once done, I would like to implement the replace method for `]]<-()` that currently allows users to easily invalidate a `QFeatures` object.

cvanderaa:
I implemented `removeAssay` and `replaceAssay`. This allowed me to easily replace the generic method `[[<-()` and hence addresses issue #57 .

I updated the list above with things that must be discussed.

cvanderaa:
The PR is ready for merge!

NEWS.md
```diff
@@ -1,5 +1,14 @@
 # QFeatures 1.7
 
+## QFeatures 1.7.2
+
+- fix: improved `addAssay()`. The improvement solves issue #104, but 
+  also dramatically improve computational time. colData is automatically
```
lgatto: - is this a `fix` or a `feat`?
- improve*s* computational tima

R/QFeatures-class.R
```diff
@@ -73,22 +74,47 @@
...
-##'   is a `character(1)` naming the single assay (default is
-##'   `"newAssay"`), and is ignored if `y` is a list of
-##'   assays. `assayLinks` is an optional [AssayLinks].
-##'
+##' - `addAssay(x, y, name, assayLinks)`: Adds one or more
+##'   new assay(s) `y` to the `QFeatures` instance `x`. `name`
+##'   is a `character(1)` naming the assay if only one assay is 
+##'   provided, and is ignored if `y` is a list of assays. `assayLinks`
+##'   is an optional [AssayLinks]. The `colData` present in `y` is 
+##'   automatically duplicated to `colData(x)` by matching sample
```
lgatto: May be change 'duplicated' to 'added' - there is some duplication, but given that it's in a different context, I don't really consider is as a bad thing (as is implied by duplication).

R/QFeatures-class.R
```diff
@@ -73,22 +74,47 @@
...
+##'   names, that is `colnames(y)`. If the samples are not present in
+##'   `x` the rows of `colData(x)` are extended to account for the
```
lgatto: `x`, the rows (comma added)

R/QFeatures-class.R
```diff
@@ -73,22 +74,47 @@
...
+##' - `addAssay(x, y, name, assayLinks)`: Adds one or more
+##'   new assay(s) `y` to the `QFeatures` instance `x`. `name`
+##'   is a `character(1)` naming the assay if only one assay is 
+##'   provided, and is ignored if `y` is a list of assays. `assayLinks`
+##'   is an optional [AssayLinks]. The `colData` present in `y` is 
+##'   automatically duplicated to `colData(x)` by matching sample
+##'   names, that is `colnames(y)`. If the samples are not present in
+##'   `x` the rows of `colData(x)` are extended to account for the 
+##'   new samples. Be aware that conflicting information between the
+##'   `colData` in `y` and the `colData` in `x` will result in an
```
lgatto: change to `colData(y)` and `colData(x)`

R/QFeatures-class.R
```diff
@@ -73,22 +74,47 @@
...
+##'   is an optional [AssayLinks]. The `colData` present in `y` is 
+##'   automatically duplicated to `colData(x)` by matching sample
+##'   names, that is `colnames(y)`. If the samples are not present in
+##'   `x` the rows of `colData(x)` are extended to account for the 
+##'   new samples. Be aware that conflicting information between the
+##'   `colData` in `y` and the `colData` in `x` will result in an 
+##'   error.
+##' - `removeAssay(x, i)`: Removes one or more assay(s) from the
+##'   `QFeatures` instance `x`. In this context, `i` is a `character()`,
+##'   `integer()` or `logical()` that indicates which assay(s) must be
```
lgatto: which assay(s) to remove.

R/QFeatures-class.R
```diff
@@ -73,22 +74,47 @@
...
+##'   removed.
+##' - `replaceAssay(x, y, i)`: Replaces one or more 
+##'   assay(s) from the `QFeatures` instance `x`. In this context, `i`
+##'   is a `character()`, `integer()` or `logical()` that indicates 
+##'   which assay(s) must be replaced. The feature links from or to
```
lgatto: to replace.

lgatto: I assume that by feature links, you mean `AssayLinks` - please be specific.

R/QFeatures-class.R
```diff
@@ -73,22 +74,47 @@
...
+##'   `integer()` or `logical()` that indicates which assay(s) must be
+##'   removed.
+##' - `replaceAssay(x, y, i)`: Replaces one or more 
+##'   assay(s) from the `QFeatures` instance `x`. In this context, `i`
+##'   is a `character()`, `integer()` or `logical()` that indicates 
+##'   which assay(s) must be replaced. The feature links from or to 
+##'   any replaced assays are automatically removed, unless the 
+##'   replacement has the same dimension names (columns and row, order
+##'   agnostic). Be aware that conflicting information between the
+##'   `colData` in `y` and the `colData` in `x` will result in an
```
lgatto: change to `colData(y)` and `colData(x)` as above.

R/QFeatures-class.R
```diff
@@ -630,7 +653,7 @@ setReplaceMethod("names", c("QFeatures", "character"),
 ##'
 ##' @param colvars A `character()` that selects column(s) in the
 ##'     `colData`.
-##' @param index The assay indicator for `SummarizedExperiment`
+##' @param index The assay indicator within a `SummarizedExperiment`
```
lgatto, line 655: Wouldn't it be better to refer to the index of the assay?

R/QFeatures-class.R
```diff
@@ -667,36 +690,268 @@ longFormat <- function(object,
...
+    
+    ## Add the assay to the ExperimentList
+    ## NOTE: we replace using the `@` slot. Although not recommended, 
+    ## this bypasses the checks of all the elements (using 
+    ## `validObject`) in the ExperimentList as this is already 
+    ## performed when building the QFeatures object and `y` is checked
+    ## at the beginning of the function. This leads to a reduction in
+    ## computational time. 
+    el <- experiments(x)
+    for(ii in names(y)) {
```
lgatto: add a space after `for`

R/QFeatures-class.R
```diff
@@ -667,36 +690,268 @@ longFormat <- function(object,
...
+        el@listData[[ii]] <- y[[ii]]
```
lgatto, line 739: Should we consider checking the validity of the new assays?

R/QFeatures-class.R
```diff
@@ -667,36 +690,268 @@ longFormat <- function(object,
...
+    cd <- .updateColData(x, y)
+    
+    ## Replace the assay to the ExperimentList
+    ## NOTE: we replace using the `@` slot. Although not recommended, 
+    ## this bypasses the checks of all the elements (using 
+    ## `validObject`) in the ExperimentList as this is already 
+    ## performed when building the QFeatures object. This leads to a 
+    ## reduction in computational time. 
+    el <- experiments(x)
+    for(ii in names(y)) {
```
lgatto: add space after `for`

R/QFeatures-class.R
```diff
@@ -667,36 +690,268 @@ longFormat <- function(object,
...
+        el@listData[[ii]] <- y[[ii]]
```
lgatto, line 795: Should we consider checking the validity of the new assays?

R/QFeatures-class.R
```diff
@@ -618,8 +641,8 @@ setReplaceMethod("names", c("QFeatures", "character"),
                      for (i in seq_len(length(x))) {
                          al <- x@assayLinks[[i]]
                          al@name  <- unname(key_vals[key_vals[, 1] == al@name, 2])
-                         if (!is.na(al@from))
-                             al@from <- unname(key_vals[key_vals[, 1] == al@from, 2])
+                         if (!all(is.na(al@from)))
+                             al@from <- unname(key_vals[key_vals[, 1] %in% al@from, 2])
```
lgatto, line 644: I see you changed `==` into `%in%` here. Does that mean that the order isn't important?

NEWS.md
cvanderaa: Actually it is a double hit with one stone, but I thought the fix was more important than the feat (cf https://github.com/rformassspectrometry/QFeatures/issues/104). I split the sentence for more clarity.

R/QFeatures-class.R
cvanderaa: Yes I fully agree, but then I would call the argument `i` instead of `index`. The latter is to comply to `MultiAssayExperiment::longFormat()`. Should we remove `index`?

R/QFeatures-class.R
cvanderaa: Yes, this is performed at line 716. Do you agree with the checking procedure?

R/QFeatures-class.R
cvanderaa: Yes, this is performed at line 783

R/QFeatures-class.R
cvanderaa: Ouch I can't remember why I changed this... I'll revert this change.

lgatto: I couldn't say without more in-depth inspection, so better double check.

R/QFeatures-class.R
lgatto: Could you add a link to `.checkAssaysToInsert()`.

cvanderaa: Sorry here it is:  https://github.com/cvanderaa/QFeatures/blob/2e314fb5c32ed393d46506affbaf7ed0b0d4b244/R/QFeatures-class.R#L865-L893

R/QFeatures-class.R
cvanderaa: Ok I remember. When an assay as multiple elements in `from`, ie it has multiple parent assays, a warning is thrown:
```r
In key_vals[, 1] == al@from :
  longer object length is not a multiple of shorter object length
```
I'll include a unit test for this so that I don't forget next time :sweat_smile:

PR #169, opened 2022-08-31
Implement a c method - solves Issue#166

cvanderaa:
(Please first merge PR #170)

In this PR I implement the `c(),QFeatures` method for combining `QFeatures` objects. I took a few liberties that require your attention:

1. I moved away from the `c(),MultiAssayExperiment` because it does not correctly handle the merging of `colData`. In this implementation, I make sure that when 2 `QFeatures` object have conflicting colData information, an error is thrown. In the other cases, the colData is merged as for `addAssay()`, taking into account the addition of new samples.
2. I only allow user to provide `QFeatures` objects. If they provide a `list` or an object that inherits from `List` or `SummarizedExperiments`, the method throws an error and suggest `addAssay()` instead. If they provide something else than those or a `QFeatures` object, the objects are converted to a `QFeatures` obbject. 
3. I allowed `QFeatures` objects to be combined with `MultiAssayExperiment` objects. This required to implement a `coerce` method from `MultiAssayExperiment` to `QFeatures`, because the generic implementation leads to an invalid (empty) `assayLinks` slot. 
4. Beside the `c(),QFeatures` method, I implement the `nrows()` and `ncols()` methods. They would be handy next to `dims()`.


cvanderaa:
I reverted the version bump when merging the current master (so it's still 1.7.2). Let me know if I need to bump the version.

lgatto (approved):

lgatto:
@cvanderaa - could you confirm that this PR can be merged.

cvanderaa:
PR is ready for merging.

PR #178, opened 2022-11-08
fix: solved bug in selectRowdata()

cvanderaa:


lgatto (approved):

lgatto:
@cvanderaa - to fix R-CMD-check-bioc, could you upate the GHA workflow definition and install MsCoreUtils version 1.11.1 directly from github.

Issue #184, opened 2023-04-16
feature request: remove empty assays


lgatto:
I don't think we have a function that removes empty assays from a QFeatures object yet - something like this:

```r
> x
An instance of class QFeatures containing 12 assays:
 [1] psmA5: SummarizedExperiment with 1 rows and 11 columns 
 [2] psmA9: SummarizedExperiment with 0 rows and 11 columns 
 [3] psmA11: SummarizedExperiment with 0 rows and 11 columns 
 ...
 [10] peptides: SummarizedExperiment with 1 rows and 33 columns 
 [11] norm_peptides: SummarizedExperiment with 1 rows and 33 columns 
 [12] proteins: SummarizedExperiment with 1 rows and 33 columns 
> nrows(x)
        psmA5         psmA9        psmA11     log_psmA5     log_psmA9 
            1             0             0             1             0 
   log_psmA11     peptideA5     peptideA9    peptideA11      peptides 
            0             1             0             0             1 
norm_peptides      proteins 
            1             1 
> nrows(x) > 0
        psmA5         psmA9        psmA11     log_psmA5     log_psmA9 
         TRUE         FALSE         FALSE          TRUE         FALSE 
   log_psmA11     peptideA5     peptideA9    peptideA11      peptides 
        FALSE          TRUE         FALSE         FALSE          TRUE 
norm_peptides      proteins 
         TRUE          TRUE 
> x[, , nrows(x) > 0]
harmonizing input:
  removing 66 sampleMap rows not in names(experiments)
An instance of class QFeatures containing 6 assays:
 [1] psmA5: SummarizedExperiment with 1 rows and 11 columns 
 [2] log_psmA5: SummarizedExperiment with 1 rows and 11 columns 
 [3] peptideA5: SummarizedExperiment with 1 rows and 11 columns 
 [4] peptides: SummarizedExperiment with 1 rows and 33 columns 
 [5] norm_peptides: SummarizedExperiment with 1 rows and 33 columns 
 [6] proteins: SummarizedExperiment with 1 rows and 33 columns 
Warning message:
'experiments' dropped; see 'metadata' 
```
I think this would be handy. 

@cvanderaa, what do you think?

cvanderaa:
Yes I agree with you, I also often use `x[, , nrows(x) > 0]` or `x[, , ncols(x) > 0]` . 

Implementing a function (eg `dropEmptyAssays()`) is straightforward, but I'm wondering what is the benefit of keeping empty assays in the first place. 

[...]

PR #185, opened 2023-04-18
feat: new dropEmptyAssays()

lgatto:
See #184


cvanderaa (commented):
Looks good to me! I see there are no unit tests. I'm wondering what would happen if `object` is empty.  Do you want me took take care of this?

lgatto:
Yes, a unit test would be great... but I haven't taken the time to write one (naughty me).

lgatto:
Addressed in #187

PR #187, opened 2023-04-20
test: added tests for dropEmptyAssays()

cvanderaa:
Note that I added one minor line to the function to fulfill an edge case.

### DIFF
--- a/R/QFeatures-class.R
+++ b/R/QFeatures-class.R
@@ -1155,7 +1155,9 @@ dropEmptyAssays <- function(object, dims = 1:2) {
         stop("Argument 'dims' must be in '1:2'.")
     if (1 %in% dims)
         object <- object[, , nrows(object) > 0]
-    if (2 %in% dims)
+    if (2 %in% dims) {
         object <- object[, , ncols(object) > 0]
+    }
+    if (!length(object)) return(QFeatures())
     object
 }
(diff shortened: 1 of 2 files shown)


R/QFeatures-class.R
```diff
@@ -1157,5 +1157,6 @@ dropEmptyAssays <- function(object, dims = 1:2) {
         object <- object[, , nrows(object) > 0]
     if (2 %in% dims)
         object <- object[, , ncols(object) > 0]
+    if (!length(object)) return(QFeatures())
```
lgatto, line 1160: Should you not start by this case?

cvanderaa: No, the edge case I want to solve happens after the subsetting. For example, suppose all assays have no features. `dropEmptyAssays()` will remove all assays and return an empty `Qfeatures` object. Without this line, the object is empty, but there is some leftover information, like metadata or colnames in the empty colData.

As a note, the function was working perfectly for `dropEmptyAssays(QFeatures())`.

cvanderaa: Btw, maybe this case could deserve a warning?

lgatto: Thanks. 

No, I wouldn't throw a warning, because it would be the normal/expected behaviour. A message at best, but I'm not even convinced that would be needed.

Issue #199, opened 2024-03-05
Tabular data input


lgatto:
- `QFeatures::readQFeatures()`
- `QFeatures::readSummarizedExperiment()`
- `scp::readSCP()`
- `scp::readSCPfromDIANN()` 

This should ideally be harmonised, possibly all moved to `QFeatures` (with more standard names), to avoid multiple re-implementations.

lgatto:
@cvanderaa - I plan to move `scp::readSCPfromDIANN()` to `QFeatures` and use the latest data in `MsDataHub` to show its working. What about renaming it `QFeatures::readQFeaturesFromDIANN()` and keep a copy as  `scp::readSCPfromDIANN()` pointing to the former's man page.

Edit: there should be one difference between  `QFeatures::readQFeaturesFromDIANN()` and `scp::readSCPfromDIANN()` though - the former produces a `QFeatures` of `SummarizedExperiment` objects, the latter of `SingleCellExperiment`, which might require some refactoring.

lgatto:
Here's a proposed hack, with minimal refactoring, to illustrate the desired behaviour (shown in the next comment). See [this comment](https://github.com/rformassspectrometry/QFeatures/issues/199#issuecomment-1982441188) for a description of what needs to be done for a proper implementation.

```r
## in QFeatures - original readSCPfromDIANN implementation
.readQFeaturesFromDIANN <-
    function (colData, reportData, extractedData = NULL,
              ecol = "MS1.Area", multiplexing = "none", ...) {
    diannReportCols <- c("File.Name", "Precursor.Id", "Modified.Sequence")
    if (!all(diannReportCols %in% colnames(reportData)))
        stop("'reportData' is not an expected DIA-NN report table ",
            "output. This function expects the main output file as ",
            "described here: https://github.com/vdemichev/DiaNN#main-output-reference")
    if (!ecol %in% colnames(reportData))
        stop("'", ecol, "' not found in 'reportData'")
    if (!"File.Name" %in% colnames(colData))
        stop("'colData' must contain a column named 'File.Name' that provides ",
            "a link to the 'File.Name' column in 'reportData'")
    if (multiplexing == "none" && !is.null(extractedData))
        stop("Providing 'extractedData' for label-free experiments ",
            "('multiplexed == \"none\"') is not expected. Raise an ",
            "issue if you need this feature: ",
            "https://github.com/UCLouvain-CBIO/scp/issues/new/choose")
    args <- list(...)
[...]

PR #200, opened 2024-03-08
Import reading functions from scp - see issue 199 for details.

lgatto:


tests/testthat/test_utils.R
```diff
@@ -0,0 +1,21 @@
+test_that(".splitSC", {
```
cvanderaa: it's .splitSE :wink:

R/QFeatures-constructors.R
```diff
@@ -1,85 +1,233 @@
-##' @title QFeatures from tabular data
+#' @title QFeatures from tabular data
```
cvanderaa: A `#` got dropped :wink:

R/QFeatures-constructors.R
```diff
@@ -1,85 +1,233 @@
...
 ##'
 ##' @description
 ##'
-##' Convert tabular data from a spreadsheet or a `data.frame` into a
-##' `QFeatures` object.
+##' These functions convert tabular data into dedicate data
```
cvanderaa: into dedicate**d**?

R/QFeatures-constructors.R
```diff
@@ -1,85 +1,233 @@
...
-##' @param ecol A `numeric` indicating the indices of the columns to
-##'     be used as expression values. Can also be a `character`
-##'     indicating the names of the columns. Caution must be taken if
-##'     the column names are composed of special characters like `(`
-##'     or `-` that will be converted to a `.` by the `read.csv`
-##'     function. If `ecol` does not match, the error message will
-##'     display the column names as seen by the `read.csv` function.
+##' - The multi-set case will generate a `QFeatures` object with
+##'   multiple [SummarizedExperiment] sets, resulting from splitting
+##'   the input table.
```
cvanderaa: Maybe we could already notify here that the multi-set case is especially interesting when `table` contains data for multiple runs/batches?

R/QFeatures-constructors.R
```diff
@@ -1,85 +1,233 @@
...
 ##'
-##' @param fnames An optional `character(1)` or `numeric(1)`
-##'     indicating the column to be used as feature names.
+##' @param assayData A `data.frame`, or any object that can be coerced
```
cvanderaa, line 21: If I'm not mistaken, the `assayData` not only holds quantitative data but also feature annotations. That's why in `readSCP()` I called it `featureData`.

I don't have a strong opinion on this, so I'll let you have the last word.

R/QFeatures-constructors.R
```diff
@@ -1,85 +1,233 @@
...
+##'     to a `data.frame`, holding the quantitative assay. For
+##'     `readSummarizedExperiment()`, this can also be a
+##'     `character(1)` to a filename.
 ##'
-##' @param ... Further arguments that can be passed on to `read.csv`
-##'     except `stringsAsFactors`, which is always `FALSE`.
+##' @param colAnnotation The type of this parameter will define
+##'     whether the resulting `QFeaures` object will contain a single
```
cvanderaa: Qfea**t**ures

tests/testthat/test_readFeatures.R
```diff
@@ -1,65 +1,166 @@
...
-    ft3 <- readQFeatures(f, ecol = 1:10, name = "psms", fname = 11)
+    ft2 <- readQFeatures(x, ecol = 1:10, name = "psms", fnames = 11)
+    ft3 <- readQFeatures(x, colAnnotation = 1:10, name = "psms", fnames = 11)
     expect_equal(ft1, ft2)
     expect_equal(ft1, ft3)
     ecol <- c("X126", "X127C", "X127N", "X128C", "X128N", "X129C",
               "X129N", "X130C", "X130N", "X131")
     ft1 <- readQFeatures(x, ecol = ecol, name = "psms")
-    ft2 <- readQFeatures(f, ecol = ecol, name = "psms")
+    ft2 <- readQFeatures(x, colAnnotation = ecol, name = "psms")
```
cvanderaa, line 25: Out of curiosity, would it work if `colAnnotation` is a vector of boolean (eg obtained after using `grepl()`). The documentation doesn't mention it, but the code seems to allow for it.

R/QFeatures-constructors.R
```diff
@@ -1,85 +1,233 @@
...
 ##'
-##' @return An instance of class [QFeatures] or [SummarizedExperiment].
+##'     For the multi-set case, a `data.frame` or any object that can
+##'     be coerced to a `data.frame`. It is expected to contain all
+##'     the sample meta information. Required fields are the
+##'     acquisition batch (given by `batchCol`) and the acquisition
+##'     channel within the batch (e.g. TMT channel, given by
+##'     `channelCol`). Additional fields (e.g. sample type,
+##'     acquisition date,...) are allowed and will be stored as sample
+##'     meta data.
```
cvanderaa: Maybe "annotation" instead of "meta data"? 

We could already mention that: "... and will be stored as sample meta data *in the `colData` of the `QFeatures` object."

R/QFeatures-constructors.R
```diff
@@ -1,85 +1,233 @@
...
-##' @param name An `character(1)` to name assay in the `QFeatures`
-##'     object. If not set, `features` is used.
+##'     For the single-set case, a `numeric` indicating the indices of
+##'     the columns to be used as expression values, or a `character`
+##'     indicating the names of the columns.
 ##'
-##' @return An instance of class [QFeatures] or [SummarizedExperiment].
+##'     For the multi-set case, a `data.frame` or any object that can
+##'     be coerced to a `data.frame`. It is expected to contain all
+##'     the sample meta information. Required fields are the
```
cvanderaa: Maybe "annotations" instead of "meta information"?

R/QFeatures-constructors.R
```diff
@@ -1,85 +1,233 @@
...
+##'     used as feature names. If missing, in the latter case, the
+##'     rows are named PSM1, PSM2, PSM3, ... Note that rownames must
+##'     be unique in `QFeatures` sets.
 ##'
-##' @describeIn readQFeatures See description.
+##' @param name For the single-set case, an optional `character(1)` to
+##'     name the set in the `QFeatures` object. If not set, `psms`
+##'     is used.
+##'
+##' @param batchCol For the multi-set case, a `numeric(1)` or
```
cvanderaa: I know the name `batchCol` comes from `readSCP`, but now that we move it to `QFeatures`, maybe it could be worth discussing/improving the nomenclature. 

I realise after many uses of `readSCP()` that `batchCol` always points to an MS run identifier (either as an MS run name or a file ID). Moreover, the word "batch" is more often used in another context and refers to a group of samples that are processed together (sample prep batch), or samples that are run on the same LC column (LC batch) or samples that are acquired sequentially within a restricted time frame (acquisition batch) rather than samples that are acquired within the same MS acquisition run (I think I'm the only one doing that :sweat_smile: ). Finally, `batchCol` is also required for LFQ experiments (when starting from PSM-level data) and in that case, calling an MS run a "batch" doesn't make sense since the run only contains a single sample. 

We could rename `batchCol` to `runCol` to avoid ambiguity.

R/QFeatures-constructors.R
```diff
@@ -1,85 +1,233 @@
...
+##'     the column name in both table are either identical and
+##'     syntactically valid (if you supply a `character`) or have the
+##'     same index (if you supply a `numeric`). Note that characters
+##'     can be converted to syntactically valid names using
+##'     `make.names`
+##'
+##' @param channelCol For the multi-set case, a `numeric(1)` or
+##'     `character(1)` pointing to the column of `colData` that
+##'     contains the column names of the quantitative data in
+##'     `assayData` (see Example).
```
cvanderaa: Maybe clarify that `channelCol` is useful for multiplexed experiments such as mTRAQ, TMT (SILAC?)

R/QFeatures-constructors.R
```diff
@@ -1,85 +1,233 @@
...
+##'     `NULL`. If both `ecol` and `colAnnotation` are set, an error
+##'     is thrown.
+##'
+##' @param verbose A `logical(1)` indicating whether the progress of
+##'     the data reading and formatting should be printed to the
+##'     console. Default is `TRUE`.
+##'
+##' @return An instance of class `QFeatures` or
+##'     [SummarizedExperiment::SummarizedExperiment()]. For the
+##'     former, the expression data of each batch is stored in a
```
cvanderaa: "expression data", I think this is still an artefact from `MSnbase` and the `ExpressionSet` class :grin:. For consistency, maybe use "assay data"  or "feature data". Also maybe use "each run" instrad of "each batch" (cf comment above).

R/QFeatures-constructors.R
```diff
@@ -1,85 +1,233 @@
...
+##' (colann <- data.frame(file = rep(paste0("File", 1:3), each = 10),
+##'                       Channel = rep(names(hlpsms)[1:10], 3)))
+##'
+##' qf3 <- readQFeatures(hlpsms, colAnnotation = colann,
+##'                      batchCol = "file", channelCol = "Channel")
+##' qf3
+##' colData(qf3)
+NULL
+
+## Simple case, with a single table to populate one QFeatures set
```
cvanderaa: What's the purpose of this line?

R/QFeatures-constructors.R
```diff
@@ -138,3 +286,239 @@ QFeatures <- function(..., assayLinks = NULL) {
...
+}
+
+.readQFeatures2 <- function(assayData, colAnnotation,
+                            batchCol, channelCol, suffix = NULL,
+                            sep = "", removeEmptyCols = FALSE,
+                            fnames, verbose = TRUE) {
+    if (missing(batchCol))
+        stop("Please provide a 'batchCol'.")
+    if (missing(channelCol))
+        stop("Please provide a 'channelCol'.")
```
cvanderaa: These checks are overkill, R will complain anyway if these arguments are missing.

R/QFeatures-constructors.R
```diff
@@ -138,3 +286,239 @@ QFeatures <- function(..., assayLinks = NULL) {
...
+    for (i in seq_along(se)) {
+        ## Add unique sample identifiers
+        colnames(se[[i]]) <- paste0(names(se)[[i]], sep, suffix)
+        ## Remove the columns that are all NA
+        if (removeEmptyCols) {
+            sel <- colSums(is.na(assay(se[[i]]))) != nrow(se[[i]])
+            se[[i]] <- se[[i]][, sel]
+        }
+    }
+    if (verbose) message("Formatting sample metadata (colData)")
```
cvanderaa: "annotations" instead of "metadata"

R/QFeatures-constructors.R
```diff
@@ -138,3 +286,239 @@ QFeatures <- function(..., assayLinks = NULL) {
...
+
+
+##' @title  Read DIA-NN output as a QFeatures objects
+##'
+##' @description
+##'
+##' This function takes the output tables from DIA-NN and converts them
+##' into a `QFeatures` object.
+##'
+##' @param colData A `data.frame` or any object that can be coerced to
```
cvanderaa: `colAnnot` instead of `colData` for consistency with `readQFeatures()`?

R/QFeatures-constructors.R
```diff
@@ -138,3 +286,239 @@ QFeatures <- function(..., assayLinks = NULL) {
...
+##'     a `data.frame`. `colData` is expected to contains all the
+##'     sample annotations. We require the table to contain a column
+##'     called `File.Name` that links to the `File.Name` in the DIA-NN
+##'     report table. If `multiplexing = "mTRAQ"`, we require a second
+##'     column called `Label` that links the label to the sample (the
+##'     labels identified by DIA-NN can be retrieved from `Modified
+##'     Sequence` column in the report table).
+##'
+##' @param reportData A `data.frame` or any object that can be coerced
```
cvanderaa: `assayData` instead of `reportData` for consistency with readQFeatures()?

R/QFeatures-constructors.R
```diff
@@ -138,3 +286,239 @@ QFeatures <- function(..., assayLinks = NULL) {
...
+##'     `"MS1.Area"`.
+##'
+##' @param multiplexing A `character(1)` indicating the type of
+##'     multiplexing used in the experiment. Provide `"none"` if the
+##'     experiment is label-free (default). Available options are:
+##'     `"mTRAQ"`.
+##'
+##' @param ... Further arguments passed to [readQFeatures()].
+##'
+##' @return An instance of class `QFeatures`. The expression data of
```
cvanderaa: "quantitative data" or "feature data" instead of "expression data"?

tests/testthat/test_readFeatures.R
lgatto: Yes, I think it should indeed work.

lgatto: Added.

R/QFeatures-constructors.R
lgatto: Comments to relate the code to the docs. I have udpated it to clarify.

R/QFeatures-constructors.R
lgatto: removed

R/QFeatures-constructors.R
lgatto: ok

R/QFeatures-constructors.R
lgatto: ok

R/QFeatures-constructors.R
lgatto: yes, indeed

lgatto:
I've addressed all comments and changes. Any idea what will break once this is merged? What about vignettes in `SCP.replication`?

cvanderaa:
After careful inspection of the new code I would like to rediscuss the some of the arguments (`colAnnot`, `ecols`, `channelCol` and `batchCol`). Currently, the documentation describes 2 cases for `readQFeatures()`:

### The single-set case

The user will use the following signature if they provide `ecols` (backward compatibility): 

```r
c("readQFeatures", c("data.frame", "missing"))
```
or better
```r
c("readQFeatures", c("data.frame", "vector"))
```
**Problem**: users cannot provide a sample annotation table in the single-set case.

## The multi-set case

The user will use the following signature

```r
c("readQFeatures", c("data.frame", "data.frame"))
```
**Problem**: "lazy" users that don't have sample annotations are stuck. One argument could be that we promote good practice and enforce users to make a sample annotation table, but then, we must enforce it as well in the single-set case. Also, it sounds a bit confusing that `colAnnot` can refer to a table of sample annotations or a vector of column name/indices, which are different concepts to me. 

## My (bold) suggestion

1. `colAnnot` should always be a table  (data.frame and friends), but the argument is optional.
2. Unify `ecols` (single-set case) and `channelCol` (multiset case) as a single argument (that I will tentatively call `ecols2` for clarity). 
     - When `ecols2` is a vector and `colAnnot` is missing: use it to select quantitative columns in  `assayData` (same as `ecols`)
     - When `ecols2` is a vector and `colAnnot` is provided: use it to select quantitative columns in  `assayData` (same as `ecols`)
     - When `ecols2` is a `character(1)` and `colAnnot` is provided: use it to select the quantitative column in  `assayData` (same as `ecols`). If no match,  use `ecols2` to match a column in `colAnnot` that provide the names of the quantitative columns (same as `channelCol`)
     - `ecols2` can **never** be missing!
3. Use `batchCol` as follows: 
    - If `batchCol` is a vector (`stopifnot(length(batchCol) == nrow(assayData))`), use that vector to split the SE. 
    - If `batchCol` is a `character(1)` and `colAnnot` is provided, use it to retrieve the sample annotation column to use for splitting the SE.
    - If `batchCol` is missing, don't split (=single-set case).

## Conclusion

I think implementing this will enable more flexible use cases, without a loss of the current capabilities.

- **Single-set case**: You can construct the `QFeatures` object with or without `colData`.
- **Multi-set case**: You can read the data without the need for a `colAnnot` table. This could facilitate reading LFQ data where a `channelCol` makes no sense. 

I will implement this, so to show my ideas in action, but I'm happy to already hear your thoughts on that.

cvanderaa:
> I've addressed all comments and changes. Any idea what will break once this is merged? What about vignettes in `SCP.replication`?

Since the behavior of the functionality did not change and there is backward compatibility, I don't think anything will break as long as we adapt `readSCP()` as a wrapper function as you suggested in #199

lgatto:
> Since the behavior of the functionality did not change and there is backward compatibility, I don't think anything will break as long as we adapt readSCP() as a wrapper function as you suggested in #199.

But some arguments have changed (for now at least), such as batchCol to runCol, which will lead to errors.

lgatto:
There are some aspects I like, others that I'm unsure to
understand. But here are a first set fo comments, and I'll think more
about your suggestion.

> Problem: users cannot provide a sample annotation table in the
> single-set case.

Yes, they can, by setting a colAnnotation that defines a single
run/batch. See the 'Single-set case using a data.frame' example.

> Also, it sounds a bit confusing that colAnnot can refer to a table
> of sample annotations or a vector of column name/indices, which are
> different concepts to me.

These are two different ways to indicate annotate columns, so not
entirely different, but still confusing I agree. My thinking was to
steer away from `ecol` and promote its replacement by a
`colAnnotation` table.

Your suggestion is the opposite, i.e. to always use `ecol`, and only
use `colAnnotation` table to add a `colData` to the final `QFeatures`. 

In your case 2,

> - When `ecols2` is a vector and `colAnnot` is provided: use it to
>   select quantitative columns in `assayData` (same as `ecols`)

I assume here `colAnnot` is used to populate the `colData` of the
returned `QFeatures` object?

lgatto:
And what about the following, that simplifies you suggestion (I
think). Here the user uses either a colAnnotation data.frame, or one
or two arguments.

## colAnnotation is missing

- The user must always provide the `ecol` vector with variables names
  or indices or a logical to identify what columns are quantitative
  data. This defines de vertical splitting of the assay and the row
  data.

- If only `ecol` of provided, and the second vector `runCol` is
  missing (or `NULL`), we are in the single-set scenario.

- If `runCol` is also provided, it must be a `numeric(1)` or a
  `character(1)` that gives the column name that defines the
  horizontal splitting. The presence of `runCol` defines the multi-set
  case.

## colAnnotation is present

- When `colAnnotation` is present, `ecol` and `runCol` must either be
  missing (or both `NULL`). `colAnnotation` must be a `data.frame` or
  an object that can be coerced to a `data.frame`.

- `colAnnotation` must at least contain a column names `ecol` that
  with the names of the samples, i.e. table column names used for
  vertical splitting of the data. This is equivalent of using the
  `ecol` argument, except it is passed in a `data.frame`.

- `colAnnotation` can also contain a `runCol` column that defines the
  horizontal splitting of the input table.

- If `colAnnotation` only contains `ecol`, we are in the single-set
  case, and `nrow(colAnnotation)` corresponds to the number of
  samples.

- If `colAnnotation` contains `ecol` and `runCol`, we are in the
  multi-set case, and `nrow(colAnnotation) == nrow(table)`. This
  situation requires a bit more work from the user, but that work
  would anyway have to be done to provide additional sample
  annotations (see next point). In addition, this is also something a
  GUI from `scpGUI` could help with.

- Any other columns of `colAnnotation` will become `colData`
  variables.

Notes: 
- [Here's a blog post](https://www.jimhester.com/post/2019-04-17-missing/) that advises to rely on `NULL` instead of missing. Will try to dig it up.
- And at this point, we could rename `ecol` to `sampleCol`, to match it with `runCol`.

cvanderaa:
I really like your suggestion and how to present the problem as vertical and horizontal splitting.

I'm however skeptic with the following:

>If colAnnotation contains ecol and runCol, we are in the
multi-set case, and nrow(colAnnotation) == nrow(table). This
situation requires a bit more work from the user, but that work
would anyway have to be done to provide additional sample
annotations (see next point). In addition, this is also something a
GUI from scpGUI could help with.

For SCP data for instance, `nrow(table)` can exceed 1,000,000 lines meaning that the `colAnnot` lines will need to be duplicated throusands of times.

Instead, I would suggest to keep nrow(`colAnnot`) equal to the number of samples in both cases (ie with or without a `runCol` column). When `colAnnot` contains a `runCol` column, use that column to retrieve the run identifiers and (thanks to an efficient internal function), match the column in `assayData` that contains the same run identifiers. This will require a bit more functionality to perform the matching of columns, but the task should be trivial.

What do you think?

cvanderaa:
> And at this point, we could rename ecol to sampleCol, to match it with runCol.

Yes I prefer `sampleCol`, but what about `quantCol`? Again, no strong opinion on this, so I let you decide.

lgatto:
I was wrong, `nrow(colAnnotation)` would need to be equal to number of samples time number of runs. Is that what you meant? In other words, in that case, `colAnnotation` would be the final `colData`.

lgatto:
> Yes I prefer `sampleCol`, but what about `quantCol`? Again, no strong opinion on this, so I let you decide.

Whichever we choose, it should be plural.

lgatto:
1. Single-set case, multiplexed: requires `quantCols` only. Also LF with a re-ordered peptide/protein-level table (runs are missing in this case).

```
|------+------------+-----------|
| cols | Quant 1..N | more cols |
|      |            |           |
|      |            |           |
|      |            |           |
|------+------------+-----------|
```

2. Multi-set case, multiplexed: requires `quantCols` and `runCol`.
```
|-----+------+------------+-----------|
| Run | cols | Quant 1..N | more cols |
|   1 |      |            |           |
|   1 |      |            |           |
|-----+------+------------+-----------|
|   2 |      |            |           |
|-----+------+------------+-----------|
```

3. Multi-set case, LF: requires `quantCols` and `runCol` with a optional multiPlexCol (for plexDIA).
```
|-----+------+---------+-----------+-------|
| Run | cols | Quant 1 | more cols | multi |
|   1 |      |         |           |       |
|   1 |      |         |           |       |
|-----+------+---------+-----------+-------|
|   2 |      |         |           |       |
|-----+------+---------+-----------+-------|
```

4. Special case DIANN. A specialised function that parses the table to case 2.

Users can either use the arguments above or a `colAnnotation` data.frame (that will become the `colData`).

cvanderaa:
I'll work tonight on refactoring `readQFeatures()` to accommodate for the new arguments and behaviour. I will provide the changes in a new PR.

PR #205, opened 2024-04-09
docs: improved docs for readQFeatures

cvanderaa:
This is a suggestion for improving the docs for `readQFeatures()`.


R/readQFeatures.R
```diff
@@ -25,12 +32,11 @@
 ##' `runCol` parameters, whether passed by the `colData` `data.frame`
 ##' and/or the `quantCols` and `runCol` arguments.
 ##'
-##' ## Single-set case
+##' @section Single-set case:
```
lgatto: Not sure we need a `@section` tag here, as we want this to be a sub-section within the Details section, not a new one.

R/readQFeatures.R
```diff
@@ -46,13 +52,11 @@
 ##' input (see next section).
 ##'
 ##'
-##' ## Multi-set case
+##' @section Multi-set case:
```
lgatto: Not sure we need a `@section` tag here, as we want this to be a sub-section within the Details section, not a new one.

R/readQFeatures.R
```diff
@@ -46,13 +52,11 @@
...
 ##'
 ##' A run/batch variable is required to import multi-set data. It can
-##' either be defined as a `runCol` argument only, or a `runCol`
-##' argument and a `runCol` variable in `colData` (see multi-set cases
-##' in the example). These cases can be represented schematically as
-##' shown below.
+##' either be defined as a `runCol` argument only. These cases can be
+##' represented schematically as shown below.
```
lgatto: Why have you removed the part with `colData`? One of the confusing things is that the same behaviour can be obtained with different sets of parameters, so this needs to be explained.

R/readQFeatures.R
```diff
@@ -11,12 +11,19 @@
 ##'
 ##' - The single-set case will generate a `QFeatures` object with a
 ##'   single `SummarizedExperiment` set containing all features of the
-##'   input table.
+##'   input table. This single-set case should be used when reading
```
lgatto: I wouldn't be so directive. Instead of *should be used*, I would suggest *are generally used*.

R/readQFeatures.R
```diff
@@ -11,12 +11,19 @@
 ##'
 ##' - The single-set case will generate a `QFeatures` object with a
 ##'   single `SummarizedExperiment` set containing all features of the
-##'   input table.
+##'   input table. This single-set case should be used when reading 
+##'   data at the peptide or at the protein level. These data are
+##'   contained in a table where the quantitative values for each 
+##'   sample are contained in a separate column.
```
lgatto: Not sure this last sentence is correct. You can have samples contained in separate columns in the for the multi-set case too.

R/readQFeatures.R
```diff
@@ -11,12 +11,19 @@
...
+##'   data at the peptide or at the protein level. These data are
+##'   contained in a table where the quantitative values for each 
+##'   sample are contained in a separate column. 
 ##'
 ##' - The multi-set case will generate a `QFeatures` object containing
 ##'   multiple `SummarizedExperiment` sets, resulting from splitting
 ##'   the input table. This multi-set case should be used when the
-##'   input table contains data from multiple runs/batches?
+##'   input table contains data from multiple runs/batches, as it is 
+##'   the case when reading data at the PSM or the precursor level.
```
lgatto: as it is generally the case when ...

R/readQFeatures.R
```diff
@@ -65,20 +69,65 @@
...
 ##' ```
-##'
-##' Whe using a `colData` `data.frame` as input, it must contain at
-##' least `quantCols` and `runCol` (in the multi-set case)
-##' variables. Additional sample annotations are desirable, and the
-##' reason such a `colData` can be passed directly to populate the
-##' `colData` slot of the returned object. When using the `quantCols`
-##' and `runCol` only, the `colData` slot contains zero
+##' 
+##' **Why splitting the table into multiple sets?** In the table
```
lgatto: May be make this a new sub-section with `##`

R/readQFeatures.R
```diff
@@ -65,20 +69,65 @@
...
+##' ensure that every sample is contained in a separate column. 
+##' However, this would assume that rows of the table are shared 
+##' across runs. When working with PSM data, every PSM is 
+##' uniquely characterised by the run it is found. Hence, there are no
+##' shared feature identifiers across runs that would enable a sound
+##' widening of the table. Splitting the table into sets is the only 
+##' solution for having every sample in a separate column while 
+##' preserving the PSM identity.
+##' 
+##' @section Adding sample annotations with `colData`:
```
lgatto: This should be a new sub-section of details with `##`

R/readQFeatures.R
```diff
@@ -65,20 +69,65 @@
...
+##' downstream data manipulation and data analyses cumbersome. A
+##' solution could be to reshape the table into a wide format to
+##' ensure that every sample is contained in a separate column. 
+##' However, this would assume that rows of the table are shared 
+##' across runs. When working with PSM data, every PSM is 
+##' uniquely characterised by the run it is found. Hence, there are no
+##' shared feature identifiers across runs that would enable a sound
+##' widening of the table. Splitting the table into sets is the only 
+##' solution for having every sample in a separate column while 
+##' preserving the PSM identity.
```
lgatto: I am not convinced by this explanation. It is correct, of course, but it makes reference to technical terms like `SummarziedExperiment`, that don't necessarily help with understanding the why. 

I am not sure I will have time to rephrase it today, but the emphasis should be on the experimental design and the data acquisition.

R/readQFeatures.R
```diff
@@ -65,20 +69,65 @@
...
+##' 
+##' We strongly recommend providing sample annotations when creating
+##' a `QFeatures` object. The `colData` is a table where each row 
+##' corresponds to a sample and each column provides information about
+##' the sample. There is no restriction on the number of columns and 
+##' to the type of data they should contain. However, we impose one or
+##' two columns (depending on the use case) that allow to link the 
+##' annotations of each sample to its quantitative data:
+##' 
+##' *Single-set case:* the `colData` must contain a column named
```
lgatto: Make this a bullet point.

R/readQFeatures.R
```diff
@@ -65,20 +69,65 @@
...
+##' `quantCols` that provides the names of the columns in `assayData`
+##' that contain the quantitative values for each sample (see 
+##' single-set cases in the examples).
+##' 
+##' *Multi-set case:* the `colData` must contain a column named
```
lgatto: Make this a bullet point.

R/readQFeatures.R
cvanderaa: my bad, reverting to title

R/readQFeatures.R
cvanderaa: I agree

R/readQFeatures.R
cvanderaa: I guess you have seen that I moved the detailed documentation of the colData in a separate "section". I think because it may be confusing, it is better to first explain the concept with the schematics and introduce the `quantCols` and `runCol` arguments, and then introduce the `colData` argument separately.

R/readQFeatures.R
cvanderaa: In the multi-set case, the quantitative values of different samples are contained in a shared column. So this sentence is correct (in the way I mean it), but may need to be clarified? Maybe "in a separate column" is not appropriate and "in an individual column" is better?

R/readQFeatures.R
cvanderaa: Of course, I can revert the paragraph structure and keep colData description as part of the use cases, but than I still think we should introduce colData after runCol and quantCols.

lgatto: I would suggest to mention the two arguments,  quantCols vector or colData data.frame, without going into details. And then, as you suggest, explain the colData approach in details.

lgatto (approved):

Issue #206, opened 2024-04-09
Create vignette for `readQFeatures()` and co


cvanderaa:
We could have a vignette devoted to reading data with readQFeatures. It would provide a bit more details on how the function works and also provides concrete examples for the different use case. For instance, we could explicitly show how to read LFQ vs labelled data, PSM vs peptide/protein data. This vignette would show how flexible and easy reading data is for any proteomics dataset. It would also serve as a useful reference everytime someones ask "can I use it for reading [name of software/technology] data?", see for example this issue: https://github.com/UCLouvain-CBIO/scp/issues/48.

I think the readSCP vignette could provide us with a head start: https://uclouvain-cbio.github.io/scp/articles/read_scp.html

lgatto:
Suggestion: start the vignette with a *Quickstart Guide* that describes 
- Importing MaxQuant data
- Importing DIA-NN data
- Importing plexDIA data
- Importing ProteomeDiscoverer data
- ...

lgatto:
This needs (subsets of) example in `MsDataHub`.

lgatto:
See [Supported input formats for readQFeatures()](https://rformassspectrometry.github.io/QFeatures/articles/readQFeatures2.html)

PR #210, opened 2024-04-24
docs: created readQFeatures vignette (issue#206)

cvanderaa:
cf #206 for background

This PR is still work in progress. 

One improvement could be to extend the section "*What about other input formats?*", where we could explicitely show different uses cases with real data fetched from `MsDataHub`.


cvanderaa:
Oh no, I pushed too fast in the wrong branch... So, here I added a commit that solves the issue in https://github.com/UCLouvain-CBIO/scp/issues/77. 

Maybe good to revive this PR.

cvanderaa:
I think this is ready for merge, but it needs your approval.

lgatto (approved):

PR #219, opened 2025-02-06
fix fnames for readQFeatures()

lgatto:


R/readQFeaturesFromDIANN.R
```diff
@@ -61,6 +61,10 @@
 ##' ## using default arguments
 ##' readQFeaturesFromDIANN(x)
 ##'
+##' ## use the precursor identifier as assay rownames
+##' readQFeaturesFromDIANN(x) |>
+##'     rownames()
```
cvanderaa, line 66: I think the example is incomplete, I would expect `fnames` to be used here.

lgatto: Indeed. Fixed.

cvanderaa (approved):
This can be merged

PR #222, opened 2025-02-08
fixes #174

### LINKED ISSUE #174: Bug in nNA()
There is a bug in `nNA()` when one assay is empty.

### Reproducible example

```r
library(QFeatures)
data("feat2")
feat2[[1]] <- feat2[[1]][, FALSE, drop = FALSE]
nNA(feat2, 1:3)
```
```
Error in h(simpleError(msg, call)) : 
  error in evaluating the argument 'args' in selecting a method for function 'do.call': different row counts implied by arguments
```

lgatto:


R/QFeatures-missing-data.R
```diff
@@ -27,9 +32,15 @@
 ## SummarizedExperiment object
 .nNAByMargin <- function(object, MARGIN = 1) {
     x <- assay(object)
-    nNA <- apply(is.na(x), MARGIN, sum)
-    n <- ifelse(MARGIN == 1, ncol(x), nrow(x))
-    pNA <- nNA / n
+    if (nrow(x) == 0 || ncol(x) == 0) {
+        nNA <- NA_integer_
+        pNA <- NA_real_
+        names(pNA) <- ""
```
cvanderaa, line 38: Is this line needed? Line 46 seems to unname the vector.

cvanderaa (approved):
Looks great, I have only a minor comment. Thanks for the fix!

R/QFeatures-missing-data.R
lgatto: pNA is expectant to have a name (see lin 44): feature name of sample name, depending. However, in this case, when there's no assay, it doesn't, hence setting it as "". 

On line 46, on the other hand, we want the value without its name.

cvanderaa: My bad, I overlooked it :+1:

cvanderaa: Quick question, why returning a DataFrame with NAs rather than with an empty row, hence  replacing `nNA <- NA_integer` with ` nNA <- integer(0)`? 

I don't have an opinion which one is best, just curious to have your thoughts :wink:

lgatto: `nNA <- integer(0)` breaks the code as it is now, so let's leave it. It's a edge case anyway.

Issue #221, opened 2025-02-07
Unique precursor identifiers


lgatto:
- A function to create precursor identifiers. DIA-NN has the
  `Precursor.Id` column, but not MaxQuant. Even if we create such an
  identifier by collating the precursor's modified sequence and its
  charge, these arent't unique (not sure why). The function could be

```r
createPrecursorId <-
    function(object, ## QFeatures
             name = "Precursor.Id", ## default name
             fcols = c("Modified.Sequence", "Precursor.Charge")) {
        ## 'name' must not be already present
        ## Adds a new column to each assays rowData
    }
```

- A function to returns TRUE or FALSE depending on the uniqueness of a
  precursor identifier. `duplicatedPrecursorId` would return FALSE for
  unique precursor identifier, and TRUE for duplicated ones. It would
  be used as shown below

```r
filterFeatures(object, ~ !duplicatedPrecursorId(Precrusor.Id))
```

lgatto:
WIP in [uniquePrecId branch](https://github.com/rformassspectrometry/QFeatures/tree/uniquePrecId) and #223

lgatto:
Some points to consider:
- [ ] When using `joinAssay()` with `fcol`, should the original names get changes in the result? Not sure we can do otherwise. 
- [ ] When calling [.setAssayRownams()](https://github.com/rformassspectrometry/QFeatures/blob/uniquePrecId/R/readQFeatures.R#L527) in `joinAssay(..., fcol)`, should all assays in the object, or only those [defined in `i`](https://github.com/rformassspectrometry/QFeatures/blob/uniquePrecId/R/readQFeatures.R#L540) be changed? 
- [ ] Currently `rownames(x[[1]]) <- ...` is possible, and the corresponding assayLinks are deleted. Do we want to disallow this? Or add a `force` to `replaceAssay()`?

PR #223, opened 2025-02-23
Create unique precursor ids and use them in joinAssays()

lgatto:
See issue #221


lgatto:
Hi @cvanderaa - this is a draft PR. This is just a first pass and unit tests are missing anyway, but it would be useful to have you input to make sure this addresses our discussions on the train back from EuBIC.

lgatto:
The best way to get started in the read `?createPrecursorId` and run the example.

cvanderaa:
This is really great to have this and will definitely streamline so of my analyses. The example is very clear and informative. Let me know when I can do an in-depth review or if you want me to contribute something (eg unit tests?).

lgatto:
I have been thinking about this. If we limit this use-case to precursors, and situations were they are nameless because fnames was not set, and or there initially was no proper names, we are OK. We can't enforce this though. If joining with an fcol is used on an assay that has links to others, the links will get lost due to row renaming (with a warning, if I'm not mistaken),

We have two options:
1. Leave it like that, and be very explicit in the manual page and highlighting the risk.
2. Recursively update the from and to slots of the assay links when renaming assays. This might not be that much work, but isn't trivial.

@cvanderaa - do you have any opinion?

R/QFeatures-join.R
```diff
@@ -149,14 +152,20 @@ mergeSElist <- function(x) {
...
+    if (!is.null(fcol)) {
+        ## Set assay rownames for joining
+        message("Using '", fcol, "' to join assays.")
+        x <- .setAssayRownames(x, fcol)
+    }
     ## Join assays and add to x
     joined_se <- mergeSElist(experiments(x)[i])
     x <- addAssay(x, joined_se, name = name)
     ## Add the multi-parent AssayLinks
     addAssayLink(x, from = i, to = name)
```
cvanderaa: I see a third option: use the `varFrom` and the `varTo` arguments of `addAssayLink()` to pass `fcol` as the linking variable between the joined sets. I think that should work, although I didn't test it.

cvanderaa:
I would like to revive this PR as I am interested by the functionality. Do you want me to take over?

lgatto:
Yes, please go ahead.

cvanderaa:
Is there a reason this has been closed without merging? I just fixed the unit tests. https://github.com/rformassspectrometry/QFeatures/pull/250

lgatto:
It should have been merged, sorry. I think I was confused with #250

cvanderaa (approved):

PR #224, opened 2025-02-25
Optimisation of aggregateFeatures

leopoldguyot:
Optimisation of `aggregateFeatures` when aggregating multiple assays.

Context: [scp Optimisation of aggregateFeaturesOverAssays #80 ](https://github.com/UCLouvain-CBIO/scp/pull/80)

I updated the docs and added a unit test to check that the variant columns are discarded.

A new benchmark is on the way.

@lgatto @cvanderaa

### DIFF
--- a/R/QFeatures-aggregation.R
+++ b/R/QFeatures-aggregation.R
@@ -1,36 +1,42 @@
-##' @title Aggregate an assay's quantitative features
+##' @title Aggregate assays' quantitative features
 ##'
 ##' @description
 ##'
-##' This function aggregates the quantitative features of an assay,
-##' applying a summarisation function (`fun`) to sets of features.
+##' This function aggregates the quantitative features of one or
+##' multiple assays, applying a summarisation function (`fun`) to
+##' sets of features.
 ##' The `fcol` variable name points to a rowData column that defines
 ##' how to group the features during aggregate. This variable can
 ##' eigher be a vector (we then refer to an *aggregation by vector*)
 ##' or an adjacency matrix (*aggregation by matrix*).
 ##'
-##' The rowData of the aggregated `SummarizedExperiment` assay
+##' The rowData of the aggregated `SummarizedExperiment` assays
 ##' contains a `.n` variable that provides the number of parent
 ##' features that were aggregated.
 ##'
 ##' When aggregating with a vector, the newly aggregated
-##' `SummarizedExperiment` assay also contains a new `aggcounts` assay
+##' `SummarizedExperiment` assays also contains a new `aggcounts` assay
 ##' containing the aggregation counts matrix, i.e. the number of
 ##' features that were aggregated for each sample, which can be
 ##' accessed with the `aggcounts()` accessor.
 ##'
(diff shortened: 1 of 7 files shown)


R/QFeatures-aggregation.R
```diff
@@ -1,36 +1,42 @@
...
 ##' @param fcol A `character(1)` naming a rowdata variable (of assay
 ##'     `i` in case of a `QFeatures`) defining how to aggregate the
-##'     features of the assay. This variable is either a `character`
+##'     features of the assays. This variable is either a `character`
 ##'     or a (possibly sparse) matrix. See below for details.
 ##'
-##' @param name A `character(1)` naming the new assay. Default is
-##'     `newAssay`. Note that the function will fail if there's
+##' @param name A `character()` naming the new assays.
+##'     name must have the same length as i.
```
lgatto: `name` must have ...

lgatto (changes requested):
I have bumped the version to 1.17.4. Could you update the NEWS entry with this change.

lgatto (approved):
I'm happy with this. @cvanderaa should also approve before merging. 

@leopoldguyot - add yourself as a contributor in the DESCRIPTION file.

cvanderaa (approved):
The PR looks good to me. Just a remark now that your PR highlights some part of the documentation. We have decided some time ago to move away from the `assay` nomenclature, keeping the word "assay" to refer to the quant data, retrieved using the `asssay()` function. Therefore, "assay" should be replaced by "set". But maybe this would be part of another issue, where we should go systematically through the whole documentation :shrug:

lgatto:
Yes, we should change this across all manual page, so let's keep assay for now.

leopoldguyot:
`❯ checking for missing documentation entries ... WARNING
  Undocumented S4 methods:
    generic '[[<-' and siglist 'QFeatures,ANY,ANY'
  All user-level objects in a package (including S4 classes and methods)
  should have documentation entries.
  See chapter ‘Writing R documentation files’ in the ‘Writing R
  Extensions’ manual.`

Do you have an idea on why the checks produced a warning for this PR and not in the master branch? The `[[<-` method is not documented in both.

lgatto:
No. Could it be that one used bioc-devel, and hence r-devel, that might have introduced additional checks?

leopoldguyot:
Should I add documentation for this method?

lgatto:
If needed, yes. An alias and a short paragraph in the main QFeature man might be enough.

man/QFeatures-class.Rd
```diff
@@ -25,7 +25,7 @@
 \alias{longFormat}
 \alias{removeAssay}
 \alias{replaceAssay}
-\alias{[[<-,QFeatures,ANY,ANY-method}
+\alias{[[<-,QFeatures,ANY,ANY,ANY-method}
```
cvanderaa: I think the issue may to stem from here, I guess when you used roxygen :shrug:

leopoldguyot: Indeed! Thanks, it should work now.

PR #229, opened 2025-03-21
Type metadata

lgatto:


lgatto:
Waiting that the relevant changes are also made in `scp` before considering any merge.

leopoldguyot:
Should .setQFeaturesType be exported ? If we want to use it in `scp::readSCP` it should be.

R/QFeatures-class.R
```diff
@@ -1175,12 +1336,62 @@ setMethod("updateObject", "QFeatures",
...
+
+
+##' @rdname QFeatures-class
+##' @param type `character(1)` that defines the type of the QFeatures.
+##'     The type can be either "bulk" or "scp" (default is "bulk"),
+##'      see [validQFeaturesType] for accepted types.
+##'
+##' @export
+##'
+setQFeaturesType <- function(object, type = "bulk") {
```
lgatto: May be not set a default `type` to force us to be explicit? What do you think @cvanderaa ?

lgatto:
This PR must be merged before similar changes in scp.

leopoldguyot:
Hello @cvanderaa I think we can start the review process. This will be followed by the review of the associated `scp` PR: [scp#90](https://github.com/UCLouvain-CBIO/scp/pull/90)

R/QFeatures-class.R
```diff
@@ -260,11 +264,15 @@ NULL
...
-                   assayLinks = "AssayLinks"),
-         prototype = prototype(
-             version = "0.3"))
+    contains = "MultiAssayExperiment",
+    slots = c(
+        version = "character",
+        assayLinks = "AssayLinks"
+    ),
+    prototype = prototype(
+        version = "0.3"
```
cvanderaa, line 272: Should we increase the version with this new change? Maybe not because the type is stored in metadata and doesn't change the class structure.

R/QFeatures-class.R
```diff
@@ -273,56 +281,74 @@ setClass("QFeatures",
...
-                  cat(" ...")
-                  cat(sprintf("\n [%i] %s: %s with %s rows and %s columns",
-                              seq(o_len)[(n-2):n], o_names[(n-2):n], elem_cl[(n-2):n],
-                              featdim[(n-2):n], sampdim[(n-2):n]), "\n")
-              }
-          })
+setMethod(
+    "show", "QFeatures",
+    function(object) {
+        type <- suppressWarnings(getQFeaturesType(object))
```
cvanderaa: Why do you need to suppress a warning? Could comment this somewhere (in the function or above)?

R/QFeatures-class.R
```diff
@@ -1175,12 +1335,84 @@ setMethod("updateObject", "QFeatures",
...
+##' @param object An instance of class [QFeatures].
+##' @param type `character(1)` defining the type of the QFeatures.
+##'     Must be one of the values returned by [validQFeaturesType()].
+##'
+##' @return 
+##' - `setQFeaturesType()`: returns the updated `QFeatures` object with
+##'   the type stored in its metadata.
+##' - `getQFeaturesType()`: returns a character string indicating the
+##'   type of the `QFeatures` object. If no type is explicitly set, 
+##'   it is inferred from the class of the experiments.
```
cvanderaa, line 1314: Maybe expand how this is inferred. If the approach is complex, you can expand on it in the details section

R/QFeatures-class.R
```diff
@@ -1175,12 +1335,84 @@ setMethod("updateObject", "QFeatures",
...
+##' - `validQFeaturesType()`: character vector of valid QFeatures types.
+##'
+##' @details
+##' These functions control an internal metadata field (`._type`) used to
```
cvanderaa: Since it is for developers only, maybe good to mention it is stored in the metadata slot?

R/QFeatures-class.R
cvanderaa: Nevermind I got it 😅 maybe still useful to have a comment

tests/testthat/test_Features.R
```diff
@@ -933,3 +933,26 @@ test_that("dropEmptyAssays", {
     )
     expect_identical(test, QFeatures())
 })
+
+test_that("setQFeaturesType", {
+    qf <- QFeatures()
+    expect_equal(metadata(qf)[["._type"]], NULL)
```
cvanderaa, line 907: Is this not violating your assumption that a QFeatures should be of a type returned by validQFeaturesType()?

tests/testthat/test_readQFeatures.R
```diff
@@ -136,24 +139,28 @@ test_that("readQFeatures: testing use cases", {
     ## Without colAnnot
     expect_identical(
         readQFeatures(x, quantCol = 1, runCol = "file"),
-        QFeatures(List(el_exp))
+        QFeatures(List(el_exp),
+                  metadata = list("._type" = "bulk"))
```
cvanderaa, line 143: I think we discussed this in person, but could you remind me why we don't want bulk as a default type?

cvanderaa (commented):
Again, looks great to me! I made a few comments. 

A more general comment that doesn't require action is that it was quite difficult for me to track the changes because there are many formatting changes along with the new implementation. It seems to have been generated automatically. I quite like the changes, but maybe double check with @lgatto whether you are using the same convention rules in your tools to avoid ping ponging between the two.

R/QFeatures-class.R
leopoldguyot: @lgatto

R/QFeatures-class.R
leopoldguyot: I will add a comment

tests/testthat/test_Features.R
leopoldguyot: I think that the goal of validQFeaturesType() is to check for valid type only if the type is set. Enforcing the need for a type would require extra development and computation work (validation etc...) and I don't think that it is what we want.

tests/testthat/test_readQFeatures.R
leopoldguyot: bulk is currently the default and the only option when using readQFeatures(). On the other hand, we choose to remove the default from setQFeaturesType() because it's a function intended for dev only and explicitly specifying the type is good practice.

leopoldguyot:
> Again, looks great to me! I made a few comments.
> 
> A more general comment that doesn't require action is that it was quite difficult for me to track the changes because there are many formatting changes along with the new implementation. It seems to have been generated automatically. I quite like the changes, but maybe double check with @lgatto whether you are using the same convention rules in your tools to avoid ping ponging between the two.

Thanks for the review! For the formatting changes, it was also a remark from @lgatto that it makes the review process more difficult and I completely agree.

I think that after these changes are approved I should create another PR to fix the formatting issues. 
While I was modifying the code I encountered a lot of inconsistencies in the tabulations (sometimes 2, sometimes 4 spaces) which can make the code difficult to read and to modify when the IDE automatically ident incorrectly. 
`styler` is a great and reliable tool to format the code. I personally use it with the Bioconductor conventions:

``` r
styler::style_pkg(transformers = biocthis::bioc_style())
```

R/QFeatures-class.R
lgatto: Yes, that's a good suggestion. Let's be explicit here.

R/QFeatures-class.R
lgatto: I would say not, as we don't change the class definition.

R/QFeatures-class.R
lgatto: Yes, the correct term for S4 classes is slots.

tests/testthat/test_Features.R
lgatto: That's a good point though. We set specific types in the `read*()` functions, because we know what the use-case and thus the type of data we have. And at the moment, we omit it this in the class constructor. I don't think (and know) what type should be set in the constructor, but this information should be explicit to the documentation.

lgatto:
Irrespective of the style (that we could discuss), a PR should address a specific point - a bug, a features, ... or code style, but not multiple things.

tests/testthat/test_Features.R
leopoldguyot: I will update the documentation of readQFeatures()

lgatto: I meant the type metadata page. Let's put all the relevant information there.

leopoldguyot: But should we also add a little sentence in the value section of readQFeatures (and readSCP)? To specify the type of the QFeatures returned by these functions?

lgatto:
I think we can say that readQFeatures() is used for bulk data, and point to
readSCP() for the single cell version, with going into any technical
details.

tests/testthat/test_readQFeatures.R
cvanderaa: I agree for the `setQFeaturesType()`, my comment was rather about the constructor, but Laurent answered the question in the thread above. 
> And at the moment, we omit it this in the class constructor. I don't think (and know) what type should be set in the constructor, but this information should be explicit to the documentation.

R/QFeatures-class.R
```diff
@@ -1175,12 +1335,84 @@ setMethod("updateObject", "QFeatures",
...
+}
+
+##' @rdname QFeatures-type
+##' @keywords internal
+##' @export
+getQFeaturesType <- function(object) {
+  stopifnot(inherits(object, "QFeatures"))
+  type <- metadata(object)[["._type"]]
+  if (is.null(type)) {
+    warning(paste("No explicit type set for this QFeatures object,",
```
lgatto: As mentioned in the other PR #236, I think a message would be more appropriate here.

R/QFeatures-class.R
```diff
@@ -1175,12 +1335,84 @@ setMethod("updateObject", "QFeatures",
...
+      type <- "bulk"
+    }
+  }
+  type
+}
+
+##' @rdname QFeatures-type
+##' @keywords internal
+##' @export
+validQFeaturesType <- function() {
```
lgatto: I think this function should be called `validQFeaturesTypes()`. See also comment in PR #236.

leopoldguyot:
Hello @lgatto could you merge the updated master into typeMetadata?

lgatto:
@cvanderaa @leopoldguyot - don't forget to approve the PR if you are happy with the changes so that it can get merged.

leopoldguyot (approved):

lgatto:
@leopoldguyot 

There's still and warning to message conversion that needs to be fixed in a unit test:

```
══ Failed tests ════════════════════════════════════════════════════════════════
── Failure ('test_Features.R:917:5'): getQFeaturesType ─────────────────────────
`type <- getQFeaturesType(qf)` did not produce any warnings.
```

leopoldguyot:
I updated the test in this PR: #242

Issue #231, opened 2025-03-21
Superfluous message in filterFeatures()


lgatto:
## Test data
 
```r
> library(QFeatures)
> data(feat2)
> rowDataNames(feat2)
CharacterList of length 3
[["assay1"]] Prot x
[["assay2"]] Prot x y
[["assay3"]] Prot x y
```

Assay 1 is missing y.

## Correct message

```r
> filterFeatures(feat2, ~ y > 1)
'y' found in 2 out of 3 assay(s)
No filter applied to the following assay(s) because one or more filtering variables are missing in the rowData: assay1.
You can control whether to remove or keep the features using the 'keep' argument (see '?filterFeature').
An instance of class QFeatures containing 3 assays:
 [1] assay1: SummarizedExperiment with 0 rows and 4 columns 
 [2] assay2: SummarizedExperiment with 2 rows and 4 columns 
 [3] assay3: SummarizedExperiment with 1 rows and 4 columns 
```

## Superfluous message

The message shouldn't be displayed anymore, or should be adapted, given that we keep the assay that is missing the filtering variable `y.

```r
> filterFeatures(feat2, ~ y > 1, keep = TRUE)
'y' found in 2 out of 3 assay(s)
No filter applied to the following assay(s) because one or more filtering variables are missing in the rowData: assay1.
You can control whether to remove or keep the features using the 'keep' argument (see '?filterFeature').
An instance of class QFeatures containing 3 assays:
 [1] assay1: SummarizedExperiment with 10 rows and 4 columns 
 [2] assay2: SummarizedExperiment with 2 rows and 4 columns 
 [3] assay3: SummarizedExperiment with 1 rows and 4 columns 
```

lgatto:
From looking at the code (not tested), it looks like it is enough to do the following in [.checkFilterVariables](https://github.com/rformassspectrometry/QFeatures/blob/master/R/QFeatures-filter.R#L371):

```r
    if (!keep & length(absent) > 0)
        msg <- c(msg, "No filter applied to the following assay(s) because ",
                 "one or more filtering variables are missing ",
                 "in the rowData: ", paste0(absent, collapse = ", "), ".\n",
                 "You can control whether to remove or keep the features ",
                 "using the 'keep' argument (see '?filterFeature').")
```

```r
> filterFeatures(feat2, ~ y > 1, keep = TRUE)
'y' found in 2 out of 3 assay(s)

An instance of class QFeatures containing 3 assays:
[...]

PR #232, opened 2025-03-21
fix issue231

lgatto:
The following PR fixes issue #231 and wraps the message the the width of the console.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: QFeatures
 Title: Quantitative features for mass spectrometry data
-Version: 1.17.6
+Version: 1.17.5
 Authors@R: c(person("Laurent", "Gatto",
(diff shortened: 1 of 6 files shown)


lgatto:
The PR is ready to me merged. The PR does the following
- Fix the bug in #231, where the message about using `keep` was still shown when the user set it.
- As part of this, the single message about  the number of assays that have the filtering variable and the 'dropped' assay (when `keep` is `FALSE`) is split in two messages, the second one only shown when `keep = FALSE` (default).
- To do this, the `keep` argument is passed to the inner `.checkFilterVariables()` function.
- The unit tests in `test_filterFeatures.R` had to be adapted, to handle the two messages.

cvanderaa:
Looks all good to me, I have no comments :+1:

cvanderaa (approved):

PR #234, opened 2025-04-23
readQFeatures(): don't ignore colData$quantCols

alyst:
I think the current version has a bug that, if there are no run information, the row names of `colData` are unconditionally assigned to the `sampleNames` (the column names of the data).
But the order of rows in colData might be different from `sampleNames`.
The correct behaviour should be to respect the `quantCols` as they are provided by the user (matching `quantCols` to `sampleNames` and reordering of colData rows happens later).

The current behaviour leads to errors in the downstream analysis (e.g. *msqrob2*), since matching data columns to the experimental design information is incorrect.

### DIFF
--- a/R/readQFeatures.R
+++ b/R/readQFeatures.R
@@ -499,7 +499,16 @@ readQFeatures <- function(assayData,
     if (is.null(colData))
         return(DataFrame(row.names = sampleNames))
     if (!length(runs)) {
-        rownames(colData) <- sampleNames
+        if ("quantCols" %in% colnames(colData)) {
+            # use quantCols column as quantCols argument
+            # can have different ordering
+            rownames(colData) <- colData$quantCols
+        } else if (length(quantCols) == nrow(colData)) {
+            # no quantCols column, we prioritize quantCols argument
+            rownames(colData) <- quantCols
+        } else {
+            rownames(colData) <- sampleNames
+        }
     } else {
         if (length(quantCols) == 1) {
             rownames(colData) <- colData$runCol
--- a/R/readQFeatures.R
+++ b/R/readQFeatures.R
@@ -499,23 +499,28 @@ readQFeatures <- function(assayData,
     if (is.null(colData))
         return(DataFrame(row.names = sampleNames))
     if (!length(runs)) {
+        # assign colData rownames to match colData to sampleNames
         if ("quantCols" %in% colnames(colData)) {
-            # use quantCols column as quantCols argument
-            # can have different ordering
+            # use colData$quantCols as rownames
+            # (quantCols argument is not guaranteed to match to colData rows)
             rownames(colData) <- colData$quantCols
         } else if (length(quantCols) == nrow(colData)) {
-            # no quantCols column, we prioritize quantCols argument
+            # no colData$quantCols column
+            # we assume the colData order matches quantCols argument
             rownames(colData) <- quantCols
         } else {
+            # assume colData order matches sampleNames
             rownames(colData) <- sampleNames
         }
     } else {
+        # run information is present, colData should match individual runs
         if (length(quantCols) == 1) {
             rownames(colData) <- colData$runCol
         } else {
             rownames(colData) <- paste0(colData$runCol, "_", colData$quantCols)
         }
     }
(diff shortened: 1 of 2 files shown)


lgatto:
Hi Alexey, thank you for your PR. Could you provide a simple example that shows the issue and then make use of it in a unit test? I am happy to write the unit test myself if you aren't familiar with `testthat`, but the small example would be useful.

alyst:
@lgatto I've added 2 unit tests with shuffled colData that fail without the fix.

lgatto:
Thank yo @alyst - we will review the PR before merging.

R/readQFeatures.R
```diff
@@ -499,14 +499,28 @@ readQFeatures <- function(assayData,
     if (is.null(colData))
         return(DataFrame(row.names = sampleNames))
     if (!length(runs)) {
-        rownames(colData) <- sampleNames
+        # assign colData rownames to match colData to sampleNames
+        if ("quantCols" %in% colnames(colData)) {
+            # use colData$quantCols as rownames
+            # (quantCols argument is not guaranteed to match to colData rows)
+            rownames(colData) <- colData$quantCols
```
cvanderaa: Thank you very much @alyst for this bug fix and for correcting it through a PR :pray: 

I was about to approve the PR, but I think the fix can be simplified. The conditional statement above should always return `TRUE` because we enforce users to provide a `quantCols` column in `colData`, see here: https://github.com/rformassspectrometry/QFeatures/blob/f745c381ace7e2b9216b01e18910dd3e44d18257/R/readQFeatures.R#L374-L378

Therefore, you could safely remove the conditional statements. The code should simplify to: 

```r
...
if (!length(runs)) {
        # assign colData rownames to match colData to sampleNames
        # use colData$quantCols as rownames
        # (quantCols argument is not guaranteed to match to colData rows)
        rownames(colData) <- colData$quantCols
} else {
...
```

If you see a use case where this is not true, could you please add it as unit test?

alyst: I also prefer this simplification. Actually, the reason why I have not implemented it -- it will make `quantCols` function argument unused. So probably it has to be removed from the function definition to further simplify the code, unless there are plans to make use of it.

cvanderaa: Good point, but I don't see a problem that `quantCols` would not used by the function, and if indeed so you could simply drop the argument from the function definition. But it is still required here: 

https://github.com/rformassspectrometry/QFeatures/blob/f745c381ace7e2b9216b01e18910dd3e44d18257/R/readQFeatures.R#L517 

However, your remark raised my attention to an important concern: `quantCols` and `colData$quantCols` may contain a different sets of entries, opening a way to data inconsistencies. This must be checked at the start of `readQFeatures()`, as part of `.checkQuantCols()`. However, I feel this is outside the scope of the PR and I will open a dedicated issue as a reminder for me to work on. 

**Conclusion**: can I ask you to do the following? 
1. Simplify the implementation above and make sure the unit tests still pass
2. Merge the current `rformassspectrometry/QFeatures` master into your branch so that the GHA checks can pass.

Many thanks in advance :pray:

alyst: @cvanderaa @lgatto I've implemented the simplifications of the `.formatColData()` logic and rebased the PR on top of the current *master*.
By looking at the overall `readQFeatures()` logic:
  * if `colData` is provided, it must contain `"quantCols"` column (checked by `.checkQuantCols()`),
  * if `runCol` is provided, `colData` must contain `"runCol"` (checked by `.checkRunCol()`).

So, regardless of `runs` presence, `.formatColData()` will execute the code paths that rely on `colData` and ignore `quantCols` argument.
That's why I have removed `quantCols` argument and the corresponding code paths from the function.
Overall, it should make the `.formatColData()` logic easier to follow.

alyst:
@cvanderaa @lgatto Could you please take a look at the PR -- I have tried to address your code review comments? Thank you!

lgatto:
Sorry for the delay. @cvanderaa - could you check and merge, please.

lgatto:
(FYI @alyst, Chris is on holidays until next week)

lgatto (approved):

PR #235, opened 2025-04-30
Type metadata (fix documentation)

leopoldguyot:
Create separate rd file for "QFeatures type" functions.


leopoldguyot:
@lgatto I think we can merge this into typeMetadata so that we can start the review process for the PR #229

PR #236, opened 2025-05-07
update QFeaturesType doc (based on review feedback)

leopoldguyot:


R/QFeatures-class.R
```diff
@@ -1353,28 +1354,41 @@ dropEmptyAssays <- function(object, dims = 1:2) {
 
 ##' Set and Get QFeatures Type
 ##'
-##' Developer-level functions to set and retrieve the type of a `QFeatures` 
+##' Developer-level functions to set and retrieve the type of a `QFeatures`
 ##' object. This type can help internal methods adapt their behavior to the
```
lgatto: Please us British spelling: behaviour.

R/QFeatures-class.R
```diff
@@ -1353,28 +1354,41 @@ dropEmptyAssays <- function(object, dims = 1:2) {
...
 ##' structure of the data.
 ##'
 ##' @param object An instance of class [QFeatures].
 ##' @param type `character(1)` defining the type of the QFeatures.
 ##'     Must be one of the values returned by [validQFeaturesType()].
```
lgatto: Change to `validQFeatureTypes()`.

lgatto: Change here and the function name below.

R/QFeatures-class.R
```diff
@@ -1353,28 +1354,41 @@ dropEmptyAssays <- function(object, dims = 1:2) {
...
 ##' - `setQFeaturesType()`: returns the updated `QFeatures` object with
 ##'   the type stored in its metadata.
 ##' - `getQFeaturesType()`: returns a character string indicating the
-##'   type of the `QFeatures` object. If no type is explicitly set, 
-##'   it is inferred from the class of the experiments.
+##'   type of the `QFeatures` object. If no type is explicitly set,
+##'   it is inferred from the class of the experiments. 
+##'   If the QFeatures contains any `SingleCellExperiment` objects, 
+##'   the type is set to "scp". Otherwise, it is set to "bulk".
 ##' - `validQFeaturesType()`: character vector of valid QFeatures types.
```
lgatto: add an *s*, as there are at least 2.

lgatto (changes requested):
Also, in `getQFeaturesType()`, in the absence of a type, wouldn't it be better to use `message()` rather than `warning()`? All old instances will return a warning, which isn't elegant.

leopoldguyot:
> Also, in `getQFeaturesType()`, in the absence of a type, wouldn't it be better to use `message()` rather than `warning()`? All old instances will return a warning, which isn't elegant.

`getQFeaturesType()` is only used internally within the show method, which calls it inside a `suppressWarnings()`.  Therefore I don't think any old instance of a `QFeatures` object would raise a warning unless the user calls `getQFeaturesType()` directly which is intended for dev only.

lgatto:
OK, point taken regarding warnings being raised with old objects. Question still stands, do we really need a warning. I don't think we need to warn users/developers when something that is expected (for old objects) happens.

R/QFeatures-class.R
```diff
@@ -1412,7 +1412,7 @@ getQFeaturesType <- function(object) {
   stopifnot(inherits(object, "QFeatures"))
   type <- metadata(object)[["._type"]]
   if (is.null(type)) {
-    warning(paste("No explicit type set for this QFeatures object,",
+    message(paste("No explicit type set for this QFeatures object,",
```
lgatto, line 1415: Don't forget to also replace `suppressWarnings()` with `suppressMessages()` in the show method.

leopoldguyot: Indeed!

PR #241, opened 2025-05-17
Long form

lgatto:


lgatto:
@leopoldguyot - please also have a look at this PR before merging.

R/QFeatures-longForm.R
```diff
@@ -0,0 +1,134 @@
...
+##' @importFrom MultiAssayExperiment longForm
+##' @importFrom reshape2 melt
+##'
+##' @importFrom BiocGenerics longForm
+##'
+##' @exportMethod longForm
+##'
+##' @aliases longForm longForm,QFeatures
+##' @aliases longFormat
+setMethod("longForm", "QFeatures",
```
leopoldguyot, line 65: Maybe it would be better to put the code of the `longForm` QFeatures method in a wrapper function like you did with the `SummarizedExperiment` method? (it will be easier to debug)

leopoldguyot (changes requested):

leopoldguyot (approved):

R/QFeatures-longForm.R
```diff
@@ -0,0 +1,139 @@
...
+}
+
+##' @noRd
+##'
+##' @export
+longFormat <- function(object,
+                       colvars = NULL,
+                       rowvars = NULL,
+                       index = 1L) {
+    .Defunct("longForm")
```
cvanderaa, line 181: Should we not start by a deprecation? This would give users time to adapt their workflow.

R/QFeatures-longForm.R
```diff
@@ -0,0 +1,139 @@
...
+##' @exportMethod longForm
+##'
+##' @aliases longForm,SummarizedExperiment
+setMethod("longForm", "SummarizedExperiment",
+          function(object, colvars = NULL,
+                   rowvars = NULL,
+                   index = seq_along(assays(object)))
+              longFormSE(object, colvars, rowvars, index, na.rm = FALSE))
+
+longFormQFeatures <- function(object, colvars = NULL,
```
cvanderaa, line 82: The longFormSE allows to remove missing values (`na.rm` argument). Is it intentional that `longFormQFeatures` doesn't?

Also both function take an `index` argument. Should this not be `i` to comply with the interface of the other functions/methods in the package?

I would personnally prefer a harmonized interface.

R/QFeatures-longForm.R
```diff
@@ -0,0 +1,139 @@
+##' @rdname QFeatures-class
+##'
+##' @param colvars A `character()` that selects column(s) in the `colData`.
+##'
+##' @param rowvars A `character()` with the names of the `rowData`
+##'     variables (columns) to retain in any assay.
+##'
+##' @param index When object is an instance of class `QFeatures`, a `numeric(1)`
```
cvanderaa: For a QFeatures, the `index` seems to allow for `numeric()`, not `numeric(1)`. Also the default is not `1L` but all assays as for `SE`s.

lgatto:
@cvanderaa - I have addressed all you comments, hopefully to your satisfaction.

cvanderaa:
Regarding the `index` and `i` naming, I realise I got confused and didn't catch it was about the assay index **within** the `SE` :man_facepalming: The documentation of `i` is now incorrect for the other functions. Do you want me to revert it? 

The rest looks good to me!

lgatto:
Yes, I can see that now. 

I have now documented `longForm()` in it's own manual page, with a reference and link in the main class manual page. I think this is the cleanest. Manual pages that are very long become too difficult to read.

lgatto:
Another point that I would appreciate your input in. Currently, we have the following discrepancy:
- `longForm,QFeatures` returns a `DataFrame`
- `longForm,SummarizedExperiment` returns a `data.frame`

`MultiAssayExperiment::longForm` returns a `DataFrame` for all methods but `MultiAssayExperiment::longForm,ANY`, that returns a `data.frame`. 

What should we do?

leopoldguyot:
> Another point that I would appreciate your input in. Currently, we have the following discrepancy:
> 
>     * `longForm,QFeatures` returns a `DataFrame`
> 
>     * `longForm,SummarizedExperiment` returns a `data.frame`
> 
> 
> `MultiAssayExperiment::longForm` returns a `DataFrame` for all methods but `MultiAssayExperiment::longForm,ANY`, that returns a `data.frame`.
> 
> What should we do?

I'd say it's better to use `DataFrame` for both cases. It's more consistent and aligns with the expectations of users working in Bioconductor workflows. 
In the case of `MultiAssayExperiment` I think they return a `data.frame` for `MultiAssayExperiment::longForm,ANY` and `MultiAssayExperiment::longForm,ExperimentList` because they assume that these implementation will only be used internally.

lgatto:
Yes, ok. 

One reason I see to use `data.frame` is to pipe it directly into tidyverse functions (the main use of `longForm()`), whereas otherwise, one needs to coerce it to a `data.frame` or a `tibble`.

lgatto:
Latest commit always returns a `DataFrame`, for consistency.

leopoldguyot:
> Yes, ok.
> 
> One reason I see to use `data.frame` is to pipe it directly into tidyverse functions (the main use of `longForm()`), whereas otherwise, one needs to coerce it to a `data.frame` or a `tibble`.

That's also a good point I didn't think of. What do you think @cvanderaa?

lgatto:
After thinking a bit more, I think we should (1) be consistent and (2) not deviate from what MAE does, hence return a `DataFrame` in both cases (which is what is currently implemented in the last commit).

cvanderaa:
I agree with both of you that `longForm` should always return a `DataFrame`. 

Good news is this (I'm currently using R 4.5):

```r 
library(S4Vectors)
library(ggplot2)
DataFrame(x = 1:10) |> 
    ggplot(aes(x = x)) + 
    geom_histogram()
```
Tadaa :tada: !! I don't know when this has been implemented.
This does not apply to `dplyr` though :sweat:

lgatto:
@cvanderaa - could you approve the PR for me to be able to merge

cvanderaa (approved):

Issue #253, opened 2026-01-09
nNA into QFeature object


lgatto:
Add a `addToObject` argument to add the col and row nNA to the QFeature's colData and rowData.

lgatto:
Related to this, I see that the example in `?nNA` isn't appropriate any more - I suspect this is since `impute()` creates a new assay, instead of replacing the one it operates on.

PR #254, opened 2026-01-19
nNA addToObject argument

leopoldguyot:
#253 for context


leopoldguyot:
I just noticed that this implementation is not robust. Indeed it stores the pNA and nNA related to samples in the global colData. The problem is that pNA and nNA are set dependent. This can be an issue when someone wants to compute missingness metrics on sets that share colnames (ie. peptide and protein level set). I can't think of a way to store the set-wise pNA and nNA without using a set-wise colData?

Issue #258, opened 2026-04-08
replaceColnames function


leopoldguyot:
## Function API
This function would look like this:

```r
replaceColnames(x = qfeatures, scol = "sample_id")
```
Where x is a `QFeatures` object and scol a `character(1)` that indicate which column of the colData will be use as colnames for the QFeatures object.

This function will change the rownames of the colData and the colnames of each set.

## Implementation

There is already a `colnames<-` method for `MultiAssayExperiments` but it expect a list of `character()` one character vector for each set.
To construct this list we will use the `colData` to have a matching between old sample names (`rownames(colData)`) and new sample names (`colData[[scol]]`). Then we will iterate on each set and construct a character vector by matching the old set colnames to the new colnames. At the end we can use the list of character and call `colnames(QFeatures) <- constructedList`.

The function should test for:
- `is.character(scol)`
- `length(scol) == 1`
- `!scol %in% colnames(colData(x))`
- `inherits(x, "QFeatures")`
- `anyDuplicated(colData(x)[[scol]])`

leopoldguyot:
@lgatto should we also have a `scol2` parameter to construct the new colnames based on the concatenation of two columns?

lgatto:
> @lgatto should we also have a `scol2` parameter to construct the new colnames based on the concatenation of two columns?

No, I think it's easy enough to do that by hand to create the colnames (why not ans `scol3`? and what `sep`? ...)

lgatto:
Ping @cvanderaa - what do you think?

PR #259, opened 2026-04-09
replaceColnames function

leopoldguyot:
cf. #258


R/QFeatures-class.R
```diff
@@ -899,10 +932,12 @@ setReplaceMethod(
...
-addAssay <- function(x,
-    y,
-    name,
-    assayLinks) {
+addAssay <- function(
+      x,
+      y,
+      name,
+      assayLinks
+) {
```
lgatto, line 940: Closing parenthesis on a new line is a big no no!

lgatto: I'll fix it myself.

lgatto (changes requested):

R/QFeatures-class.R
leopoldguyot: I think it was autoformated by `styler::style_file("path", transformers = biocthis::bioc_style())`

Issue #262, opened 2026-06-18
Function to aggregate technical replicates


EminAraftpoor:
Hi,

I was looking into the QFeatures reference manual for a function to aggregate technical replicates together but could not find one. Does a function like this exist? I saw here on GitHub that it was discussed in the past in issues like #188 and #215.

lgatto:
@leopoldguyot - could you add this feature? Let's discuss tomorrow morning.

lgatto:
@EminAraftpoor - we have tested and used it, but not added it to the package yet. Will do.

lgatto:
We will add a new `aggregateSamples()` function. 

But don't forget that in many situations, explicitly modelling replicates might be the best approach.

EminAraftpoor:
Hi Laurent,

Thank you and the rest of the team for implementing this feature. I'm looking forward to the next release.
I'm aware that generally it makes sense to account for technical replicates using mixed effects models and/or blocking but it's not applicable for the dataset I've been given.
I would like to ask for your opinion when regarding the order of operations:

1. When meaning, would you generally advise to take the arithmetic or geometric mean? In this case I have duplicates so the median and mean are interchangeable.
2. Would you recommend aggregating after batch correction/normalisation?

Looking forwards to reading your thoughts.

Best,
Emin

lgatto:
This is now available in the latest version on Github and on Bioc devel soon.

Issue #215, opened 2024-05-14
add a function to average columns of a SummarizedExperiment


jorainer:
In some cases we might want to average (technical) replicates (columns) of an `SummarizedExperiment`. Why a `SummarizedExperiment`:
- in addition to the abundance matrix it provides also information on samples
- it is a container for the preprocessing results - averaging replicates during preprocessing does not make much sense. 

The function (method?) should average all assays and also update the `colData` after reducing/combining the data of the technical replicates.

A template for the function could be the [`averageSE()`](https://github.com/EuracBiomedicalResearch/CompMetaboTools/blob/master/R/averageSE.R) function, but I would change the name (maybe to `averageColumns()`?) and maybe see if we could improve it a bit? documentation needs to be definitely improved...

Also, to avoid adding the `SummarizedExperiment` package as a dependency to *MetaboCoreUtils* we should rather use `requireNamespace()` and call the `SummarizedExperiment`-specific functions with `SummarizedExperiment::`.

lgatto:
This is also something that is in the browing in `QFeatures`. We already have `aggregateFeatures()` for rows, but want something similar for pseudo-bulking. 

- The name `averageSE()` isn't a good fit IMHO, way to general: it would bette to make an explicit reference to columns, and there might be other options that averaging. 
- In `QFeatures`, we will implement methods for `SummarizedExperiment` and `QFeatures` objects. 
- It is important to keep a consistent API (see [here](https://github.com/rformassspectrometry/QFeatures/issues/214), something we should discuss more widely for RforMS) - in this case, make sure the signature/arguments are consistent with `aggregateFeatures()`. 
- There's also probably some code-reuse and refactoring opportunities with `aggregateFeatures()`

I would very strongly suggest to add this to `QFeatures`. Any low-level function, that operates on the matrix could end up in `MsCoreUtils`.

Also adding @cvanderaa to the discussion.

cvanderaa:
Indeed, you can find an initial discussion on the matter here: https://github.com/rformassspectrometry/QFeatures/issues/188

jorainer:
yeeees, just discussed also with @philouail - and we agreed that it would be better to have that in a more high-level package instead of `MetaboCoreUtils`! So 100% for adding this to `QFeatures`. 

Also, I totally agree that we need to change the name. Maybe `aggregateColumns()` to use a similar naming?

Would I would like from this method/function:

- allow to specify which function to use for aggregating (`mean`, `median` etc)
[...]

Issue #188, opened 2023-05-18
Feature request: aggregate by sample (i.e. column)


csdaw:
Hi there,

I'm not sure if this functionality already exists, but I'd like to propose a function `aggregateSamples()` which is companion to `aggregateFeatures()`, and would do essentially the same thing but would combine columns instead of rows.
I'm currently using this to calculate log2 SILAC ratios, but it would have other general uses such as calculating row means.

Thanks!

### Functions

``` r
suppressPackageStartupMessages({
  library(SummarizedExperiment)
  library(QFeatures)
})

# Define 'column-wise' version of MsCoreUtils::aggregate_by_vector()
aggregate_cols_by_vector <- function(x, INDEX, FUN, ...) {
  ## Input checks go here
  
  FUN <- match.fun(FUN)
  
  res <- lapply(
    split(seq_len(ncol(x)), INDEX),
    FUN = function(j) FUN(x[, j, drop = FALSE], ...)
  )
  nms <- names(res)
  res <- do.call(cbind, res)
  colnames(res) <- nms
  rownames(res) <- rownames(x)
  
  ## HDF5Matrix check goes here
  res
}

# .aggregateSamples, essentally a 'column-wise' version of .aggregateQFeatures
.aggregateSamples <- function(object, scol, fun, ...) {
  if (missing(scol))
    stop("'scol' is required")
  m <- assay(object, 1)
  cd <- colData(object)
  if (!scol %in% names(cd))
    stop("'scol' not found in the assay's colData.")
  groupBy <- cd[[scol]]
  
  ## Store class of assay i in case it is not a SummarizedExperiment
  ## so that the aggregated assay can be reverted to that class
  .class <- class(object)
  
  ## Message about NA values in quant/col data goes here...
  
  if (is.vector(groupBy)) {
    aggregated_assay <- aggregate_cols_by_vector(m, groupBy, fun, ...)
    aggcount_assay <- aggregate_cols_by_vector(m, groupBy, MatrixGenerics::rowCounts)
    aggregated_coldata <- QFeatures::reduceDataFrame(cd, cd[[scol]],
                                                     simplify = TRUE,
                                                     drop = TRUE,
                                                     count = TRUE)
    assays <- SimpleList(assay = aggregated_assay, aggcounts = aggcount_assay)
    coldata <- aggregated_coldata[colnames(aggregated_assay), , drop = FALSE]
  } else {
    stop("'scol' must be a vector")
  }
[...]

PR #263, opened 2026-07-13
Implementation of aggregateSamples

leopoldguyot:
Implementation of `aggregateSamples` function. It allows to aggregates the columns of multiples sets of one QFeatures. Methods available for SE or QFeatures.

Related issues: #262 #188 #215


leopoldguyot:
There is one major issue:
When aggregating several sets at the same time, conflicting coldata data would need to be added to the global colData.
For instance let say we want to aggregate several sets by patients and cell types. For each set the colData columns that are invariants are kept, this means for instance that a run column (this column should contain the same value for all the samples of a set) will be kept. After that when we try to add the new aggregated sets to the QFeatures, the `addAssay` function will add the information of the colData of the new SummarizedExperiment to the global `colData`. But if a row is already present in the global colData and that the new set's colData contain clashing data, an error will be thrown. In our example, for one pair of patient/cell type, the run column will contain different values between the different new set (therefore throwing an error).

I can think of different way to fix the problem:
- Only allow the use of the function on a single set (`i = character(1)`)
- Extract the colData of each new Set and remove columns with clashing information before adding them to the QFeatures. (That would add a lot of extra data manipulation)
- Don't add individual colData to the SummarizedExperiment but modify the global colData directly.

The last option means that we cannot simply use the SummarizedExperiment method in the QFeatures method, we need to have a new function that apply the aggregation on a SE but without returning a colData.

What do you think @lgatto ?

NEWS.md
```diff
@@ -1,3 +1,9 @@
+# QFeatures 1.25
+
+## QFeatures 1.25.1
```
lgatto: The current devel version is 1.23.1 (see https://bioconductor.org/packages/devel/bioc/html/QFeatures.html). You need to create 1.23.2.

lgatto: Please also adapt the version in DESCRIPTION.

R/QFeatures-aggregation.R
```diff
@@ -505,3 +505,160 @@ validAdjacencyMatrix <- function(x) {
...
+##' For `QFeatures` objects, the aggregated assay `colData` is added to
+##' the global `colData` using [addAssay()]. If an aggregated group name
+##' already exists as a row name in the global `colData`, it is reused
+##' only when shared `colData` variables do not contain conflicting
+##' values, otherwise an error is thrown. The aggregation of samples should
+##' take place after run joining.
+##'
+##' @param object An instance of class [QFeatures] or [SummarizedExperiment].
+##'
+##' @param i A `character(1)` naming the assay that will be aggregated.
```
lgatto: Do you mean an assay in a SE or a set in a QFeatures?

R/QFeatures-aggregation.R
```diff
@@ -505,3 +505,160 @@ validAdjacencyMatrix <- function(x) {
...
+##'
+##' @param scol A `character()` naming the `colData` variable defining
+##'     how to group samples.
+##'     When multiple variables are supplied, their values are concatenated
+##'     with `_` to build the grouping variable.
+##'
+##' @param name A `character(1)` naming the new assay.
```
lgatto: As above, assay or set?

R/QFeatures-aggregation.R
```diff
@@ -505,3 +505,160 @@ validAdjacencyMatrix <- function(x) {
...
+##'
+##' @param fun A function used to aggregate the samples. The function should
+##'   apply its summarization row-wise.
+##'
+##' @param moreFun A named `list()` of additional aggregation functions
```
lgatto: Why this? Is it not easier/cleaner to call `aggregateSamples()` multiple times?

R/QFeatures-aggregation.R
```diff
@@ -505,3 +505,160 @@ validAdjacencyMatrix <- function(x) {
...
+##'     used to create additional assays. The function should apply its
+##'     summarization row-wise.
+##'
+##' @param ... Additional parameters passed to `fun`.
+##'
+##' @return A `QFeatures` object with additional assays or a
```
lgatto: set?

lgatto (changes requested):

leopoldguyot:
In the new 8.1.0 roxygen [release](https://github.com/r-lib/roxygen2/releases/tag/v8.1.0) there is now a "support" for `setAs`. 

https://github.com/r-lib/roxygen2/pull/1909 :
> roxygen2 didn't recognise setAs() or setGroupGeneric(), so a roxygen block in front of either produced an Rd file with no \usage — which current R-devel reports as Rd files without \usage. setAs() also got no method alias. Both now behave like setMethod() and setGeneric():
> 
> \name{coerce,numeric,Account-method}
> \alias{coerce,numeric,Account-method}
> \usage{
> \S4method{coerce}{numeric,Account}(from, to = "Account", strict = TRUE)
> }

The problem is that this add this new entry in the QFeatures-class.Rd file:
`\S4method{coerce}{MultiAssayExperiment,QFeatures}(from, to = "QFeatures", strict = TRUE)`

Which introduces new parameters that were not used in the setAs definition (`to` and `strict`). This triggers warning during checks:

> ❯ checking Rd \usage sections ... WARNING
>   Undocumented arguments in Rd file 'QFeatures-class.Rd'
>     ‘from’ ‘to’ ‘strict’
>   
>   Functions with \usage entries need to have the appropriate \alias
>   entries, and all their arguments documented.
>   The \usage entries must correspond to syntactically valid R code.
>   See chapter ‘Writing R documentation files’ in the ‘Writing R
>   Extensions’ manual.

Should I add these parameters to the setAs method or only document them @lgatto ?

R/QFeatures-aggregation.R
leopoldguyot: Changed to fixed extra assays (sd and count)

lgatto (approved):

